PR #1, opened 2015-06-03
remove $that = $this usage

samsonasik:


vaclavvanik:
Hi, you fotgot to bump php version

samsonasik:
done ;)

vaclavvanik:
:+1: 
btw. in your other PR you should do bump php again :-)

samsonasik:
done ;). I will check other components

Warm regards,

Abdul Malik Ikhsan

Pada 4 Jun 2015, pukul 12.07, Václav Vaník <notifications@github.com> menulis:

weierophinney:
:+1: This is a nice bonus of using PHP 5.5. :)

pierredup:
Changing the minimum required PHP version in a patch release is bad for BC, because this breaks when updating from 2.5.1 to 2.5.2 on a server that doesn't have PHP 5.5 installed yet.
Shouldn't breaking changes like this only happen in a major (or even minor, but not patch) release?

weierophinney:
@pierredup Starting with 2.5.0 we both messaged a bump to PHP 5.5, and stopped testing against earlier PHP versions; our mistake was that while we intended to change the PHP minimum requirement to 5.5 in the `composer.json` for 2.5.0, we only did so in the main zf2 repo, and not the component repos.

Since we *do* only test against 5.5 and up, claiming that we support less than that is also untrue. As such, we've bumped the version in the 2.5.1 version of each component to reflect what we actually support, and what we originally intended.

pierredup:
@weierophinney Thanks for the explanation

PR #3, opened 2015-06-04
Fix undefined index persistent_id on RedisResourceManager

pensiero:
Probably a typo

### DIFF
--- a/src/Storage/Adapter/RedisResourceManager.php
+++ b/src/Storage/Adapter/RedisResourceManager.php
@@ -264,7 +264,7 @@ class RedisResourceManager
         $redis  = $resource['resource'];
         if ($resource['persistent_id'] !== '') {
             //connect or reuse persistent connection
-            $success = $redis->pconnect($server['host'], $server['port'], $server['timeout'], $server['persistent_id']);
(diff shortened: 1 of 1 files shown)


Ocramius:
Why isn't this caught by tests? Maybe missing coverage?

pensiero:
Test cover only the possibility that is set a `persistent_id` or a `persistend_id` ([docs](https://github.com/zendframework/zend-cache/blob/master/test/Storage/Adapter/RedisResourceManagerTest.php#L119)) and that is found after been setted, but there are no test of the persistent_id about the `connect` method of `RedisResourceManager` ([connect](https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/RedisResourceManager.php#L261))

pensiero:
Is it possible to merge this?

marc-mabe:
@pensiero As you explained yourself this wasn't covered by a test -> please add a test case for it.

pensiero:
I can try doing it in the next days, but meanwhile is it important to fix this typo that is very clear

pensiero:
Moved to #9

PR #5, opened 2015-06-05
Set ttl to specific cache key for Redis Storage Adapter

pensiero:
Implemented two methods inside the `Storage\Adapter\Redis` that allow to:
- set a ttl to a specific key after setting it up (`setTimeout`)
- get the remaining tll of a specific key (`getRemainingTimeout`)

Actually is possible to specify a ttl to the adapter, but it will be assumed by all the keys that are setted up after the initialization.
Tests and interface implemented.

### DIFF
--- a/src/Storage/ExpirableInterface.php
+++ b/src/Storage/ExpirableInterface.php
@@ -9,6 +9,13 @@
 
 namespace Zend\Cache\Storage;
 
+/**
+ * Interface ExpirableInterface
+ * Adapters that implements this interfaces are able to set a specific ttl to their items
+ *  and get the remaining ttl applied to their items
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
 interface ExpirableInterface
 {
     /**
--- a/src/Storage/ExpirableInterface.php
+++ b/src/Storage/ExpirableInterface.php
@@ -9,32 +9,35 @@
 
 namespace Zend\Cache\Storage;
 
+use Zend\Cache\Exception;
+
 /**
  * Interface ExpirableInterface
  * Adapters that implements this interfaces are able to set a specific ttl to their items
  *  and get the remaining ttl applied to their items
  *
- * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ * @author Oscar Fanelli <oscar.nesis@gmail.com>
  */
 interface ExpirableInterface
 {
     /**
      * Sets an expiration date (a timeout) on an item
      *
-     * @param $key
-     * @param $ttl
+     * @param string $key
+     * @param int $ttl
      *
      * @return bool
+     * @throws Exception\RuntimeException
      */
     public function setTimeout($key, $ttl);
 
     /**
      * Remaining timeout of an item
      *
-     * @param $key
+     * @param string $key
      *
-     * @return mixed
+     * @return int
+     * @throws Exception\RuntimeException
      */
     public function getRemainingTimeout($key);
-
 }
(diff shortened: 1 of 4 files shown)


Maks3w:
Please rebase against current master

gianarb:
3f33aaf What fix? :dart:

pensiero:
:angry: it was something from another pull request.. damn xD i'll rebase

src/Storage/ExpirableInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Cache\Storage;
+
+interface ExpirableInterface
```
Maks3w: Please add a docblock describing the purpose of the interface. Like an introduction paragraph

pensiero:
How can I do it? The master branch of my fork contain that commit (that is used for this PR: #3)

src/Storage/ExpirableInterface.php
```diff
@@ -0,0 +1,40 @@
...
+ *  and get the remaining ttl applied to their items
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+interface ExpirableInterface
+{
+    /**
+     * Sets an expiration date (a timeout) on an item
+     *
+     * @param $key
```
Maks3w: Improve params and return types

Maks3w:
Rebase or undo the cosmetic changes. Try to keep your commits focused in the issue and don't change unrelated lines.

pensiero:
@Maks3w how can I do it?

pensiero:
PS: sorry for that, it was not intentional :(

pensiero:
Is it ok now?

marc-mabe:
1.) What is the difference of `touchItem` and `setTimeout`. It's currently not implemented in the redis adapter and falls back to https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/AbstractAdapter.php#L1081
2.) `getRemainingTimeout` should be part of `getMetadata`

pensiero:
1) `touchItem` reset the lifetime of an Item, while `setTimeout` allows to specify a different lifetime than the global one (specificied using `setTtl`)
2) How can I achieve this? I saw that ìnternalGetMetadata` return an empty array (https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/AbstractAdapter.php#L591-L598) ... is it correct?

marc-mabe:
> 1) touchItem reset the lifetime of an Item, while setTimeout allows to specify a different lifetime than the global one (specificied using setTtl)

The way to go is the following if you have to change the TTL to something different as the current:
```php
$cache->getOptions()->setTtl($ttl);
$cache->touchItem($key);
```
So you should overwrite `internalTouchItem` and set the TTL from options

> 2) How can I achieve this? I saw that ìnternalGetMetadata` return an empty array (https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/AbstractAdapter.php#L591-L598) ... is it correct?

Like it's done on MongoDB:
https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/MongoDb.php#L261-L266
https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/MongoDb.php#L241-L243

pensiero:
Moved to #10

PR #6, opened 2015-06-05
Hotfix/composer suggestions

Maks3w:


gianarb:
:+1:

composer.json
```diff
@@ -25,13 +25,15 @@
         "phpunit/PHPUnit": "~4.0"
     },
     "suggest": {
-        "zendframework/zend-serializer": "Zend\\Serializer component",
```
marc-mabe: The serializer should be a suggestion only as it's only used by the serializer plugin.

PR #7, opened 2015-06-05
[travis-ci] Adjust tests available on Travis-CI platform

Maks3w:


Maks3w:
Coverage increased (+17.18%) to 61.94%

gianarb:
:+1:

Maks3w:
Last commit remove version constraints in apc extension because is expected to be used with APCu instead APC (Deprecated since PHP 5.4)

Maks3w:
Coverage increased (+23.15%) to 67.9%

marc-mabe:
> Last commit remove version constraints in apc extension because is expected to be used with APCu instead APC (Deprecated since PHP 5.4)
What about replacing the APC adapter with a new one exactly for APCu that works well if APCu isn't in compatible mode

Maks3w:
@marc-mabe Actual APC adapter is not failling with APCu

marc-mabe:
@Maks3w I know as I did make it compatible ;) But if you compile APCu without full compatibility support there are no functions `apc_*` as all functions begin with `apcu_` -> see https://github.com/krakjoe/apcu/blob/db5d3db91470fdf07f6d873cba3f75326079869f/config.m4#L76

PR #9, opened 2015-06-13
Fix undefined index persistent_id on RedisResourceManager

pensiero:
In the RedisResourceManager there was a reference to a wrong variable (`$server` instead of `$resource`).

Added also some tests to catch the error during the test of the persistent id.

Refer to #3 , I have created a new PR to clean my code from other local branches that was creating troubles.

### DIFF
--- a/test/Storage/Adapter/RedisResourceManagerTest.php
+++ b/test/Storage/Adapter/RedisResourceManagerTest.php
@@ -33,6 +33,14 @@ class RedisResourceManagerTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * Test get the Redis resource
+     */
+    public function testGetResource()
+    {
+
+    }
+
+    /**
      * @group 6495
      */
     public function testSetServerWithPasswordInUri()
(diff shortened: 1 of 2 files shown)


marc-mabe:
looks good to me

Maks3w:
Travis-CI fails. I prefer to see #7 merged before

pensiero:
@Maks3w yes, the `assertInstanceOf` is probably related to what you have done with in #7

test/Storage/Adapter/RedisResourceManagerTest.php
```diff
@@ -111,6 +111,7 @@ public function testValidPersistentId()
         $expectedPersistentId = '1234';
         $this->resourceManager->setResource($resourceId, $resource);
         $this->assertSame($expectedPersistentId, $this->resourceManager->getPersistentId($resourceId));
+        $this->assertInstanceOf('Redis', $this->resourceManager->getResource($resourceId));
```
Maks3w: Add 

```
if (!extension_loaded('redis')) {
            $this->markTestSkipped("Redis extension is not loaded");
        }
```

To both tests

pensiero: Exactly before the line 114? https://github.com/zendframework/zend-cache/pull/9/files#diff-3fabf23591b6ac55b8f56fc86ff85597R114
Or at the beginning of the test?

Maks3w: At the beginning of the tests

pensiero: Done, I added also the check of the `TESTS_ZEND_CACHE_REDIS_ENABLED` env

marc-mabe:
:+1:

PR #10, opened 2015-06-13
Set ttl to specific cache key for Redis Storage Adapter

pensiero:
- Implemented the `internalTouchItem` method inside `Storage\Adapter\Redis` that allow to touch a Redis item.
- Implemented the `_ttl` metadata, that return the remaining tll of a specific key

Thanks to @marc-mabe to help me understanding what to do!

Refer to #5, I have created a new PR to clean my code from other local branches that was creating troubles.

### DIFF
--- a/src/Storage/Adapter/Redis.php
+++ b/src/Storage/Adapter/Redis.php
@@ -301,6 +301,25 @@ class Redis extends AbstractAdapter implements
     }
 
     /**
+     * Internal method to touch an item.
+     *
+     * @param string &$normalizedKey Key which will be touched
+     *
+     * @return bool
+     * @throws Exception\RuntimeException
+     */
+    protected function internalTouchItem(& $normalizedKey)
+    {
+        $redis = $this->getRedisResource();
+        try {
+            $ttl = $this->getOptions()->getTtl();
+            return (bool) $redis->expire($this->namespacePrefix . $normalizedKey, $ttl);
+        } catch (RedisResourceException $e) {
+            throw new Exception\RuntimeException($redis->getLastError(), $e->getCode(), $e);
+        }
+    }
+
+    /**
      * Internal method to remove an item.
      *
      * @param string &$normalizedKey Key which will be removed
@@ -426,7 +445,11 @@ class Redis extends AbstractAdapter implements
     {
         if ($this->capabilities === null) {
             $this->capabilityMarker = new stdClass();
-            $minTtl = $this->resourceManager->getMajorVersion($this->resourceId) < 2 ? 0 : 1;
+
+            $redisVersion = $this->resourceManager->getMajorVersion($this->resourceId);
+            $minTtl = version_compare($redisVersion, '2', '<') ? 0 : 1;
+            $supportedMetadata = version_compare($redisVersion, '2', '>=') ? ['ttl'] : [];
+
             //without serialization redis supports only strings for simple
             //get/set methods
             $this->capabilities     = new Capabilities(
@@ -443,7 +466,7 @@ class Redis extends AbstractAdapter implements
                         'object'   => false,
                         'resource' => false,
                     ],
-                    'supportedMetadata'  => [],
+                    'supportedMetadata'  => $supportedMetadata,
                     'minTtl'             => $minTtl,
                     'maxTtl'             => 0,
                     'staticTtl'          => true,
@@ -457,5 +480,70 @@ class Redis extends AbstractAdapter implements
         }
 
         return $this->capabilities;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @throws Exception\ExceptionInterface
+     */
+    protected function internalGetMetadata(& $normalizedKey)
+    {
+        $redis    = $this->getRedisResource();
+        $metadata = [];
+
+        try {
+            $redisVersion = $this->resourceManager->getMajorVersion($this->resourceId);
+
+            // redis >= 2.8
+            // The command 'pttl' returns -2 if the item does not exist
+            // and -1 if the item has no associated expire
+            if (version_compare($redisVersion, '2.8',  '>=')) {
+                $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
+                if ($pttl <= -2) {
+                    return false;
+                }
+                $metadata['ttl'] = ($pttl == -1) ? null : $pttl / 1000;
+
+            // redis >= 2.6
+            // The command 'pttl' returns -1 if the item does not exist or the item as no associated expire
+            } elseif (version_compare($redisVersion, '2.6', '>=')) {
+                $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
+                if ($pttl <= -1) {
+                    if (!$this->internalHasItem($normalizedKey)) {
+                        return false;
+                    }
+                    $metadata['ttl'] = null;
+                } else {
+                    $metadata['ttl'] = $pttl / 1000;
+                }
+
+            // redis >= 2
+            // The command 'pttl' is not supported but 'ttl'
+            // The command 'ttl' returns 0 if the item does not exist same as if the item is going to be expired
+            // NOTE: In case of ttl=0 we return false because the item is going to be expired in a very near future
+            //       and then doesn't exist any more
+            } elseif (version_compare($redisVersion, '2', '>=')) {
+                $ttl = $redis->ttl($this->namespacePrefix . $normalizedKey);
+                if ($ttl <= -1) {
+                    if (!$this->internalHasItem($normalizedKey)) {
+                        return false;
+                    }
+                    $metadata['ttl'] = null;
+                } else {
+                    $metadata['ttl'] = $ttl;
+                }
+
+            // redis < 2
+            // The commands 'pttl' and 'ttl' are not supported
+            // but item existence have to be checked
+            } elseif (!$this->internalHasItem($normalizedKey)) {
+                return false;
+            }
+        } catch (RedisResourceException $e) {
+            throw new Exception\RuntimeException($redis->getLastError(), $e->getCode(), $e);
+        }
+
+        return $metadata;
     }
 }
--- a/test/Storage/Adapter/RedisTest.php
+++ b/test/Storage/Adapter/RedisTest.php
@@ -312,4 +312,20 @@ class RedisTest extends CommonAdapterTest
         $this->_options->setPassword($password);
         $this->assertEquals($password, $this->_options->getPassword(), 'Password was set incorrectly using RedisOptions');
     }
+
+    public function testTouchItem()
+    {
+        $key = 'key';
+
+        // no TTL
+        $this->_storage->getOptions()->setTtl(0);
+        $this->_storage->setItem($key, 'val');
+        $this->assertEquals(0, $this->_storage->getMetadata($key)['ttl']);
+
+        // touch with a specific TTL will add this TTL
+        $ttl = 1000;
+        $this->_storage->getOptions()->setTtl($ttl);
+        $this->assertTrue($this->_storage->touchItem($key));
+        $this->assertEquals($ttl, ceil($this->_storage->getMetadata($key)['ttl']));
+    }
 }


src/Storage/Adapter/Redis.php
```diff
@@ -458,4 +479,34 @@ protected function internalGetCapabilities()
...
+
+        $redis = $this->getRedisResource();
+        try {
+
+            /**
+             * Remaining ttl of an item, in seconds
+             * In redis 2.6, the command return -1 if the key does not exist
+             * In redis > 2.8, the command return -2 if the key does not exist, -1 if the key exists but has no associated expire
+             */
+            $remainingTimeout = $redis->ttl($this->namespacePrefix . $normalizedKey);
```
marc-mabe: Be be compatible with other adapters `getMetadata` should return `false` in the case the item doesn't exist.
-> from `StorageInterface::getMetadata()` : `@return array|bool Metadata on success, false on failure` where a failure means the request can't be made but it's not an error. An error would result in an exception.

src/Storage/Adapter/Redis.php
```diff
@@ -458,4 +479,34 @@ protected function internalGetCapabilities()
...
+
+        } catch (RedisResourceException $e) {
+            throw new Exception\RuntimeException($redis->getLastError(), $e->getCode(), $e);
+        }
+
+        return [
+            '_remainingTimeout' => $remainingTimeout,
```
marc-mabe: From Redis docu:

> **ttl, pttl**
> 
> Description: Returns the time to live left for a given key in seconds (ttl), or milliseconds (pttl).

I would prefer to name this metadata `ttl` but receive the more accurate `pttl` and convert this value into a second based float to be the same as the APC adapter.

test/Storage/Adapter/RedisTest.php
```diff
@@ -312,4 +312,17 @@ public function testOptionsGetSetPassword()
...
         $this->assertEquals($password, $this->_options->getPassword(), 'Password was set incorrectly using RedisOptions');
     }
+
+    public function testTouchItem()
+    {
+        $ttl = 1000;
+        $key = 'key';
+
+        $this->_storage->setItem($key, 'val');
+        $this->assertLessThanOrEqual(0, $this->_storage->getMetadata($key)['_remainingTimeout']);
```
marc-mabe: Please set the TTL option before this line for better understanding.

test/Storage/Adapter/RedisTest.php
```diff
@@ -312,4 +312,17 @@ public function testOptionsGetSetPassword()
...
+
+        $this->_storage->getOptions()->setTtl($ttl);
+        $this->assertTrue($this->_storage->touchItem($key));
+        $this->assertEquals($ttl, $this->_storage->touchItem($key));
```
marc-mabe: Why `touchItem()` returns a TTL on second call?

src/Storage/Adapter/Redis.php
pensiero: I'm returning false if the the item is not found (https://github.com/zendframework/zend-cache/pull/10/files#diff-8799c8aa6a52387ce614380b5db83f76R491)

Did you mean that I have to update the docblock?

test/Storage/Adapter/RedisTest.php
pensiero: Mistake, it was returning true, equal to 1000. Fixed

src/Storage/Adapter/Redis.php
```diff
@@ -458,4 +479,35 @@ protected function internalGetCapabilities()
...
+             */
+            $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
+            $ttl = $pttl > 0 ? $pttl / 1000 : $pttl;
+
+        } catch (RedisResourceException $e) {
+            throw new Exception\RuntimeException($redis->getLastError(), $e->getCode(), $e);
+        }
+
+        return [
+            '_ttl' => $ttl,
```
marc-mabe: There is no need to have an underscore in front of the metadata.

src/Storage/Adapter/Redis.php
```diff
@@ -458,4 +479,35 @@ protected function internalGetCapabilities()
...
+        $redis = $this->getRedisResource();
+        try {
+
+            /**
+             * Remaining ttl of an item, in seconds
+             * In redis 2.6, the command return -1 if the key does not exist
+             * In redis > 2.8, the command return -2 if the key does not exist, -1 if the key exists but has no associated expire
+             */
+            $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
+            $ttl = $pttl > 0 ? $pttl / 1000 : $pttl;
```
marc-mabe: From you comment redis will returns `-1` if the item doesn't exist (or `-2` since redis 2.8). In this case you have to return `false` and in case of redis > 2.8 and the item does not have associated expire it should be `null`.

Because the `expire()` call already checks for a non existing item you don't need to check this before.

src/Storage/Adapter/Redis.php
pensiero: I reflected the metadata definition in the MongoDb adapter: https://github.com/pensiero/zend-cache/blob/feature/set-ttl-redis-adapter/src/Storage/Adapter/MongoDb.php#L242

test/Storage/Adapter/RedisTest.php
```diff
@@ -312,4 +312,17 @@ public function testOptionsGetSetPassword()
...
     }
+
+    public function testTouchItem()
+    {
+        $ttl = 1000;
+        $key = 'key';
+
+        $this->_storage->setItem($key, 'val');
+        $this->_storage->getOptions()->setTtl($ttl);
+        $this->assertLessThanOrEqual(0, $this->_storage->getMetadata($key)['_ttl']);
```
marc-mabe: I don't really understand this. The item stored should become the configured TTL but therefore the TTL have to be set before `setItem` and then the metadata `ttl` should return the remaining TTL.

Can you explain why in your case the TTL have to be `<= 0` please?

src/Storage/Adapter/Redis.php
marc-mabe: The `_id` in mongodb is a special one and don't need to be compatible to other adapters as It is the internal id of mongodb and is named `_id` in mongodb, too.

test/Storage/Adapter/RedisTest.php
pensiero: When we set the item, there is not TTL.
If there is no TTL set, than the metadata will be 0 (now i'm changing to `null` after your previous comment).
So, even if we set the TTL before getting the metadata, the metadata will be updated only after invoking `touchItem`.

After the `touchItem`, because we set before the TTL, the metadata will return the TTL.

src/Storage/Adapter/Redis.php
pensiero: Ok :+1:

src/Storage/Adapter/Redis.php
```diff
@@ -458,4 +479,58 @@ protected function internalGetCapabilities()
...
+                    default: $ttl = $pttl / 1000;
+                }
+            }
+            else if ($majorVersion >= 2) {
+                $ttl = $redis->ttl($this->namespacePrefix . $normalizedKey);
+                $ttl = $ttl > 0 ? $ttl : null;
+            }
+            else {
+                $ttl = false;
+            }
```
marc-mabe: This looks better but now you return an array with `ttl=false` in this case instead of `return false` and the check for `if (!$this->internalHasItem($normalizedKey)) { return false; }` can be removed.

Also I see that redis < 2 doesn't support this feature so this have to be reflect in `getCapabilities()`, too.
Only on redis < 2 you have to check for item existence:

I also don't see a difference of TTL = 0 and expired. It looks like the 0 means the item exist but have a remaining TTL of 0 seconds so it will be expire in the very near future but I didn't test it and couldn't find information quickly.

This is a fast coded version I think of but needs testing.
```php
protected function internalGetMetadata(& $normalizedKey)
{
    $redis    = $this->getRedisResource();
    $metadata = [];

    try {
        $redisVersion = $this->resourceManager->getMajorVersion($this->resourceId);

        // redis >= 2.8
        // The command 'pttl' returns -2 if the item does not exist
        // and -1 if the item has no associated expire
        if (version_compare($redisVersion, '2.8',  '>=') {
            $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
            if ($pttl <= -2) {
                return false;
            }
            $metadata['ttl'] = ($pttl == -1) ? null : $pttl / 1000;

        // redis >= 2.6
        // The command 'pttl' returns -1 if the item does not exist
        } elseif (version_compare($redisVersion, '2.6', '>=') {
            if ($pttl <= -1) {
                return false;
            }
            $metadata['ttl'] = $pttl / 1000;

        // redis >= 2
        // The command 'pttl' is not supported but 'ttl'
        // The command 'ttl' returns 0 if the item does not exist same as if the item is going to be expired
        // NOTE: In case of ttl=0 we return false because the item is goning to be expired in a very near future
        //       and then doesn't exist any more
        } elseif (version_compare($redisVersion, '2', '>=')) {
            $ttl = $redis->ttl($this->namespacePrefix . $normalizedKey)
            if ($ttl <= 0) {
                return false;
            }
            $metadata['ttl'] = $ttl;

        // redis < 2
        // The commands 'pttl' and 'ttl' are not supported
        // but item existence have to be checked
        } elseif (!$this->internalHasItem($normalizedKey)) {
            return false;
        }

    } catch (RedisResourceException $e) {
        throw new Exception\RuntimeException($redis->getLastError(), $e->getCode(), $e);
    }

    return $metadata;
}
```

As you see on redis < 2 this method isn't supported and will not be present in the returned metadata array so the redis version have to be checked also on get capabilities.

pensiero: Very nice!! Thanks!

test/Storage/Adapter/RedisTest.php
```diff
@@ -312,4 +312,17 @@ public function testOptionsGetSetPassword()
...
+    {
+        $ttl = 1000;
+        $key = 'key';
+
+        $this->_storage->setItem($key, 'val');
+        $this->_storage->getOptions()->setTtl($ttl);
+        $this->assertNull($this->_storage->getMetadata($key)['ttl']);
+
+        $this->assertTrue($this->_storage->touchItem($key));
+        $this->assertEquals($ttl, ceil($this->_storage->getMetadata($key)['ttl']));
```
marc-mabe: What I mean with the first line of `setTtl` was that you expect a specific TTL set (in this case none)  and test against this value but the TTL will not be set by the test. You only knows that this will be the default but a default can change.

So please:
```php
$key = 'key';

// no TTL
$this->_storage->getOptions()->setTtl(0);
$this->_storage->setItem($key, 'val');
$this->assertNull($this->_storage->getMetadata($key)['ttl']);

// touch with a specific TTL will add this TTL
$ttl = 1000;
$this->_storage->getOptions()->setTtl($ttl);
$this->assertTrue($this->_storage->touchItem($key));
$this->assertEquals($ttl, ceil($this->_storage->getMetadata($key)['ttl']));
```

marc-mabe:
Only the one comment above.

Please let run unit tests for this adapter and check code coverage for your new code. I know the version specific lines will only be tested if you run the specific version but one of this version should be well tested and all the other new code around.

thanks!

test/Storage/Adapter/RedisTest.php
pensiero: :+1:

pensiero:
@marc-mabe I can't do also the test for specific versions.. if someone can do that, ok

marc-mabe:
@pensiero It's very ok if you test only your version I think

pensiero:
@marc-mabe then yes! :+1:

marc-mabe:
@pensiero but some tests failed - see continuous-integration/travis-ci/pr

pensiero:
@marc-mabe it's strange, because in my environment the test pass... and I'm using php 5.6, that fails in Travis. Could it be for some configs? I've enabled `TESTS_ZEND_CACHE_REDIS_ENABLED` and started the redis server...anything else?

Maks3w:
Could you rebase against current master? Current master have many improvements for integration tests

pensiero:
@Maks3w done, but however test fails

marc-mabe:
@pensiero If I look @ https://coveralls.io/builds/2883342/source?filename=src%2FStorage%2FAdapter%2FRedis.php I see the tests ran with a redis version >= 2.0 and < 2.6 with the logic:
```php
// ...

// redis >= 2
// The command 'pttl' is not supported but 'ttl'
// The command 'ttl' returns 0 if the item does not exist same as if the item is going to be expired
// NOTE: In case of ttl=0 we return false because the item is going to be expired in a very near future
//       and then doesn't exist any more
} elseif (version_compare($redisVersion, '2', '>=')) {
    $ttl = $redis->ttl($this->namespacePrefix . $normalizedKey);
    if ($ttl <= 0) {
        return false;
    }
    $metadata['ttl'] = $ttl;
// ...
```

One of the tests failing looks like this:
```php
public function testGetMetadata()
{
    $capabilities = $this->_storage->getCapabilities();
    $supportedMetadatas = $capabilities->getSupportedMetadata();
    $this->assertTrue($this->_storage->setItem('key', 'value'));
    $metadata = $this->_storage->getMetadata('key');
    $this->assertInternalType('array', $metadata); // <- failes
    // ...
}
```

As of the test doesn't set a specific TTL it will be the default one for infinite lifetime (TTL=0). I could think of that redis could return ttl=0 in this case, too. So a returned ttl=0 could would mean one of the following:

* item exists with infinite TTL
* item exists but item will expire in the very near future (no second based TTL left)
* item doesn't exist

If this is the case we have to check for `$this->internalHasItem($normalizedKey)` as well:
```php
    if ($ttl <= 0 && !$this->internalHasItem($normalizedKey)) {
        return false;
    }
```

marc-mabe:
PS: you did not rebased against master. You have merged the master into your branch.
Please use `git fetch {remote} && git rebase {remote}/master` ;)

pensiero:
Ok we are doing progress. Now I'm returning null if the item has no associated expire (whatever the version is) and false if the items doesn't exist.

However `testGetMetadata` in [CommonAdapterTest](https://github.com/pensiero/zend-cache/blob/feature/set-ttl-redis-adapter/test/Storage/Adapter/CommonAdapterTest.php#L359) is expecting an array, even if this is empty. @marc-mabe do you think I can replace the `null` return with an empty array?

marc-mabe:
@pensiero From interface:
```php
    /**
     * Get metadata of an item.
     *
     * @param  string $key
     * @return array|bool Metadata on success, false on failure
     * @throws \Zend\Cache\Exception\ExceptionInterface
     */
    public function getMetadata($key);
```

pensiero:
@marc-mabe if as you said the tests are running with redis version >= 2 and < 2.6, how can this return a `$metadata['ttl']` of `-1` (instead of expected `0`), basing on [this case](https://github.com/pensiero/zend-cache/blob/feature/set-ttl-redis-adapter/src/Storage/Adapter/Redis.php#L525-L533)? thanks

marc-mabe:
@pensiero In this case you should return an array with `ttl` set to `NULL` instead of returning `null`

pensiero:
@marc-mabe it's what I did: https://github.com/pensiero/zend-cache/blob/feature/set-ttl-redis-adapter/src/Storage/Adapter/Redis.php#L531
but it seems that it's not the problem (or maybe I did not understand what you said)

marc-mabe:
```php
                    // ...
                    $metadata['ttl'] = null;
                }
                $metadata['ttl'] = $ttl;
                // ...
```
It's not working this way as the raw TTL returned by redis will overwrite the `ttl` 2 lines later ;)

pensiero:
WTF! :grimacing: let's see if now they pass...

pensiero:
It's seems ok, what do you think @marc-mabe ?

marc-mabe:
Looks better now but there is the same issue with redis version >= 2.6 - see https://github.com/zendframework/zend-cache/pull/10/files#diff-8799c8aa6a52387ce614380b5db83f76R516

And please check your coding style - see https://travis-ci.org/zendframework/zend-cache/jobs/68730134

Again - please make a rebase with current master and combine all your commits into a single one
```
git fetch <upstream>
git rebase -i {upstream}/master
```

pensiero:
So nice rebasing! First time used! Thanks a lot @marc-mabe (never end to learning, very happy of this)

marc-mabe:
I think it looks good now :+1: 
@Maks3w ping

pensiero:
great! thanks a lot marc!

pensiero:
Is there anything else I can do?

Issue #12, opened 2015-06-16
Why the resource is not public accessible?


pensiero:
Why the method [getRedisResource()](https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/Redis.php#L81) is protected and not public?

In my case, I need to obtain the Redis resource and pass it to some modules (in this case, [DoctrineORM](http://doctrine-orm.readthedocs.org/en/latest/reference/caching.html#redis)).

weierophinney:
It's protected likely as it was only intended for internal purposes.

That said, you have 3 options:

- Extend the class and make the method public:
```php
public function getRedisResource()
{
    return parent::getRedisResource();
}
```
  If you go this route, you'll also need to provide a custom factory to ensure your own adapter is used.
- Use reflection to make the method accessible, and then call it:
```php
$r = new ReflectionMethod($adapter, 'getRedisResource');
$r->setAccessible(true);
$resource = $r->invoke($adapter);
 ```
- Send a pull request that changes the visibility of the method.

pensiero:
Thanks a lot @weierophinney, I'll send a PR and meanwhile I use the second method :+1:

pensiero:
@weierophinney do you think it would be nice to change to public the same method in each adapter? (Redis/Memcached/Apc...) ?

weierophinney:
Yes; it's useful to be able to share the adapters, and we cannot assume
everything is configured under our own service manager.

On Tue, Jun 16, 2015 at 9:52 AM, Oscar Fanelli <notifications@github.com>
wrote:

> @weierophinney <https://github.com/weierophinney> do you think it would
> be nice to change to public the same method in each adapter?
> (Redis/Memcached/Apc...) ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-cache/issues/12#issuecomment-112459807>
> .
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

marc-mabe:
@pensiero @weierophinney The resources of all resource based storages **are** already public accessible from the resource manager. `$storage->getOptions()->getResourceManager()->getResource()`

The resource manager is where you configure and share resources.

pensiero:
@marc-mabe yes, but you can't access to the resource manager from the adapter

marc-mabe:
You can as seen in my example above. The $storage variable holds the adapter instance.

pensiero:
[...]

PR #13, opened 2015-06-16
Changed to public all the getXXXResource methods of the cache adapters

pensiero:
This PR is a consequence of the issue #12

It's useful in case someone wants to obtain the resource and pass it to other modules that need the original resource, (avoid creating and configuring a new resource, but re-use the one previously configured).

### DIFF
--- a/test/Storage/Adapter/RedisTest.php
+++ b/test/Storage/Adapter/RedisTest.php
@@ -14,18 +14,6 @@ use Redis as RedisResource;
 
 class RedisTest extends CommonAdapterTest
 {
-    /**
-     *
-     * @var Cache\Storage\Adapter\RedisOptions
-     */
-    protected $_options;
-
-    /**
-     *
-     * @var Cache\Storage\Adapter\Redis
-     */
-    protected $_storage;
-
     public function setUp()
     {
         if (!getenv('TESTS_ZEND_CACHE_REDIS_ENABLED')) {
@@ -311,5 +299,10 @@ class RedisTest extends CommonAdapterTest
         $password = 'my-secret';
         $this->_options->setPassword($password);
         $this->assertEquals($password, $this->_options->getPassword(), 'Password was set incorrectly using RedisOptions');
+    }
+
+    public function testGetRedisResource()
+    {
+        $this->assertInstanceOf('Redis', $this->_storage->getRedisResource());
     }
 }
--- a/src/Storage/Adapter/Memcache.php
+++ b/src/Storage/Adapter/Memcache.php
@@ -77,7 +77,7 @@ class Memcache extends AbstractAdapter implements
      *
      * @return MemcacheResource
      */
-    protected function getMemcacheResource()
+    public function getMemcacheResource()
     {
         if ($this->initialized) {
             return $this->resourceManager->getResource($this->resourceId);
--- a/src/Storage/Adapter/Memcached.php
+++ b/src/Storage/Adapter/Memcached.php
@@ -89,7 +89,7 @@ class Memcached extends AbstractAdapter implements
      *
      * @return MemcachedResource
      */
-    protected function getMemcachedResource()
+    public function getMemcachedResource()
     {
         if (!$this->initialized) {
             $options = $this->getOptions();
--- a/src/Storage/Adapter/MongoDb.php
+++ b/src/Storage/Adapter/MongoDb.php
@@ -75,7 +75,7 @@ class MongoDb extends AbstractAdapter implements FlushableInterface
      *
      * @return MongoResource
      */
-    private function getMongoDbResource()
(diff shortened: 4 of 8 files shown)


Maks3w:
Because this will be a public API we need specific tests for those methods.

Maks3w:
Should we create a common method name `getResource` and for v3 add it to the interface?

pensiero:
I thought about it, but there are some adapters, like the [WinCache](https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/WinCache.php) adapter, that doesn't have a `getWinCacheResource` method

weierophinney:
> Should we create a common method name getResource and for v3 add it to the interface?

The changes as stated are to support #12, and allow getting at the underlying resource for an adapter only; they are completely BC. It's even arguable as to whether this would warrant a new minor version; since only visibility is changed, the only "new" feature is that you can access the methods via instances now.

I'm not entirely sure we would want to add a `getResource()` method to the adapter interface in the future. There may be cases where the adapter might be virtual (e.g., unit test mocking) or the underlying functionality does not use a resource (e.g., file system, black hole, memory and session adapters), which would make the method semantics ambiguous. You typically want the resource because you want to pass it to another object, and in such a case, you need to know exactly what you're retrieving; an `instanceof` check followed by calling the more specific method makes sense in those situations.

Maks3w:
1. IMO it's a feature because new methods are available in public API. (What a feature mainly do)

2. I suggest to use getResource when possible as a name rule so those adapters share the same API methods.

Maks3w:
@weierophinney This PR need test for ensure public API is respected in future releases

weierophinney:
> I suggest to use getResource when possible as a name rule so those adapters share the same API methods.

@Maks3w — My point is that many of the existing adapters are not backed by PHP resources. And of those that are, some are true PHP resources, some are object instances. Additionally, documenting the return value in the interface would be tricky as a result; we could only specify `mixed`, which is not a great fit for an interface.

Since the method does not fit all adapters, and because the return value cannot be specified universally, it should *not* be in the interface.

> This PR need test for ensure public API is respected in future releases

Agreed. @pensiero — can you add tests for these, please? They can be as simple as just testing that the return value of calling the method on an instance returns something.

pensiero:
@weierophinney ok, only that the return value is not null, or also that the return value is an instance of Redis/Memcached...?

weierophinney:
@pensiero Check that it's an instance of the expected type (no need to also test `NotNull`, as an `InstanceOf` check validates that anyways).

marc-mabe:
As already sayed in #12 the resources of all resource based storages **are** already public accessible from the resource manager. 
```php
$options  = $storage->getOptions();
$resource = $options->getResourceManager()->getResource($options->getResourceId());
```
The resource manager is where you configure and share resources.

Because of the resource manager part is a little bit hard to understand I would like to see this changed for ZF3 so the storage adapter will get injected with the real resource. But the resource manager contains lot of adapter specific code to instantiate resources from given configuration. This logic needs to be somewhere but in my opinion this could be better handled in the service manager.

What do you think?

pensiero:
Do I have to do something to make this PR mergeable?

test/Storage/Adapter/RedisTest.php
```diff
@@ -14,18 +14,6 @@
 
 class RedisTest extends CommonAdapterTest
 {
-    /**
-     *
-     * @var Cache\Storage\Adapter\RedisOptions
-     */
-    protected $_options;
```
Maks3w: Don't remove this. Probably they are for to provide useful typehints for typechecker tools

Maks3w:
Please rebase. Seems to have merge conflicts.

test/Storage/Adapter/RedisTest.php
pensiero: Done, did you mean for both `$_options` and `$_storage`, isn't it?

pensiero:
Updated and resolved conflicts

marc-mabe:
I don't know if it is a good idea to proxy this method. For which reason this is needed and why only proxy this one method and not methods like `RedisResourceManger::getMajorVersion()`. In the end you could proxy all methods but than you can also implement all directly within the adapter as it would no longer make sense but it will grow up the complexity a lot.

So what's really wrong with the following ?
```
$options  = $storage->getOptions();
$resource = $options->getResourceManager()->getResource($options->getResourceId());
```

pensiero:
If it works, ok for me

pensiero:
@weierophinney it's ok also for you not to merge this and use the solution that @Maks3w proposed?

marc-mabe:
@weierophinney @Maks3w can I close this as `wontfix`

weierophinney:
Absolutely.

PR #14, opened 2015-06-19
persistent_id as string instead of int in tests of RedisResourceManager

### LINKED ISSUE #4: Discordance of Redis persistent_id typology between docs and tests
`persistent_id` of Redis adapter, [according to docs](http://framework.zend.com/manual/current/en/modules/zend.cache.storage.adapter.html#the-redis-adapter), should be something like `RDB` or `AOF`; however, in the [tests](https://github.com/zendframework/zend-cache/blob/master/test/Storage/Adapter/RedisResourceManagerTest.php#L102), the persistent_id that is setted is a generic number.

What is correct, and what is wrong?

pensiero:
Fixes #4

### DIFF
--- a/test/Storage/Adapter/RedisResourceManagerTest.php
+++ b/test/Storage/Adapter/RedisResourceManagerTest.php
@@ -57,7 +57,7 @@ class RedisResourceManagerTest extends \PHPUnit_Framework_TestCase
         $server      = 'redis://dummyuser:dummypass@testhost:1234';
         $dummyResId2 = '12345678901';
         $resource    = [
-            'persistent_id' => 1234,
+            'persistent_id' => 'my_connection_name',
             'server'        => $server,
             'password'      => 'abcd1234'
         ];
@@ -80,7 +80,7 @@ class RedisResourceManagerTest extends \PHPUnit_Framework_TestCase
         $server2     = 'redis://dummyuser:dummypass@testhost2:1234';
         $dummyResId2 = '12345678901';
         $resource    = [
-            'persistent_id' => 1234,
+            'persistent_id' => 'my_connection_name',
             'server'        => $server,
             'password'      => 'abcd1234'
         ];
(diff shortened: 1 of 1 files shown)


marc-mabe:
:+1

test/Storage/Adapter/RedisResourceManagerTest.php
```diff
@@ -57,7 +57,7 @@ public function testSetServerWithPasswordInParameters()
         $server      = 'redis://dummyuser:dummypass@testhost:1234';
         $dummyResId2 = '12345678901';
         $resource    = [
-            'persistent_id' => 1234,
+            'persistent_id' => 'my_connection_name',
```
Maks3w: Why the old value is not valid?

Maks3w:
I don't understand what this changes try to fix. Because there is no previous test failure

marc-mabe:
@Maks3w It's only for clarification because of tests are often used to see how a system have to be used and at this point you could be thinking wrang that the `persistent_id` have to be an integer but it doesn't. On the same time the documentation was is wrong, too but in a very different way. see zendframework/zf2-documentation#1495

Maks3w:
Then an expection is expected to be trown if value is invalid?

marc-mabe:
@Maks3w An integer isn't invalid it's simply an integer passed an function it expects a string so this will be automatically converted to a string. It will be invalid in PHP-7 only with `strict_types=1`

pensiero:
Do you think it could be merged?

Maks3w:
There's nothing in the PR to prevent revert the change. May throw  Exception if type is invalid?

pensiero:
Sorry @Maks3w but I didn't understand

Maks3w:
If previous value was wrong. Why test was working?. Should we add a check and throw exception if the value has the wrong type?

Anyway I rely on @marc-mabe to decide if this should be merged.

pensiero:
As @marc-mabe said in his last post, there's nothing wrong, because there is automatic conversion from `int` to `string`, but this way it's more clear that it's not an *integer id*, but a string.

marc-mabe:
@Maks3w the test works because it's not totally wrong. It simply passes an integer value to something that expects a string and the integer will be internally converted to a string.
It's like testing 'substr(1234,1,1)'.

But as sad before test are often used so see how to interact with a lib and with this in mind it's better to "fix" it as someone could wrongly think he needs to pass an integer value instead of a string.

pensiero:
Is it possible to merge so I can close the PR? Thanks

pensiero:
Thanks :)

PR #15, opened 2015-06-23
[BugFix] Using memcached on HHVM

rms230:
This fix is specificly for HHVM, when adding servers there is an undefined index error.

### DIFF
--- a/src/Storage/Adapter/MemcachedResourceManager.php
+++ b/src/Storage/Adapter/MemcachedResourceManager.php
@@ -160,7 +160,7 @@ class MemcachedResourceManager
         // merge and add servers (with persistence id servers could be added already)
         $servers = array_udiff($resource['servers'], $memc->getServerList(), [$this, 'compareServers']);
         if ($servers) {
-            $memc->addServers($servers);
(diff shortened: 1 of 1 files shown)


Maks3w:
I think this needs a more in depth analisys.

* There are more occurrences of `addServers($servers)`
* May this should be done in the return of `$resource->getServerList()`

Maks3w:
@rms230 I'll close this due inactivity.

gianarb:
@Maks3w this closure's motivation is an internal role of @zendframework/community-review-team team? :)
Because into the ZF2 repo there are a lot of "inactive pr" :+1:

weierophinney:
Re-opening. This corrects a number of issues on HHVM, bringing us closer to having the component work on that platform.

Maks3w:
@weierophinney Seriously?

Maks3w:
> I think this needs a more in depth analisys.

There are more occurrences of addServers($servers)
May this should be done in the return of $resource->getServerList()

weierophinney:
@Maks3w It's a start, which is far better than not taking it at all and leaving it completely uncorrected.

Maks3w:
rm -rf P2P

PR #16, opened 2015-08-05
Update composer.json

remicollet:
Serializer is in the "require"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
         "phpunit/PHPUnit": "~4.0"
     },
     "suggest": {
-        "zendframework/zend-serializer": "Zend\\Serializer component",
         "zendframework/zend-session": "Zend\\Session component",
         "ext-apcu": "APCU, to use the APC storage adapter",
         "ext-dba": "DBA, to use the DBA storage adapter",


gianarb:
:+1:

marc-mabe:
I habe  a dejavu :)

The serializer is only needed for Thema serializer plugin and is not required for the hole component to work.
-> it should be in suggest only!

weierophinney:
Agreed with @marc-mabe ; anything that is optional and not part of the default use case should be in "suggest" (as it currently is).

If you need that component for your _application_ use case, you should add it to your own dependencies, per the suggestion.

remicollet:
@weierophinney if it is optional, please remove it from "require", current file is inconsistent.

gianarb:
@weierophinney @marc-mabe maybe you have read bad this PR.. The problem is that `zendframework/zend-serializer` is suggested and required.. it's inconsistent IMO..

marc-mabe:
@gianarb  zendframework/zf2#2960

gianarb:
I don't understand.. Why zend-serializer is suggested and required!?

marc-mabe:
@gianarb It should not be both that's indeed wrong. It should be 'suggest' only (and 'require-dev' for testing) as @remicollet did in #17

PR #17, opened 2015-08-05
Cleanup composer.json

remicollet:
Serializer in is "suggest" since 99f1ee0876b1c65bfdd26228919287318fd2a4a7 and is not mandatory according to closed PR #16
So make dep consistent.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,6 @@
         "php": ">=5.5",
         "zendframework/zend-stdlib": "~2.5",
         "zendframework/zend-servicemanager": "~2.5",
-        "zendframework/zend-serializer": "~2.5",
         "zendframework/zend-eventmanager": "~2.5"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "zendframework/zend-eventmanager": "~2.5"
     },
     "require-dev": {
+        "zendframework/zend-serializer": "~2.5", 
         "zendframework/zend-session": "~2.5",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"


gianarb:
If I'm using zend-cache without other components this pull request is very uncomfortable IMO

remicollet:
@gianarb I don't really care if it is mandtory or not, I just like to fix the composer.json inconsistency ;) 
(and the other way, PR#16 have be refused)

gianarb:
o.O fix for fix eheh

marc-mabe:
@remicollet :+1

Maks3w:
@marc-mabe This is BC Break

marc-mabe:
In which case(s) breaks that BC - and what to do than ?

Maks3w:
Anyone which does not have zend-serializer in his composer.json will get a PHP Error because Zend\Serializer classes not exists.

marc-mabe:
... mh you are right but only if it not loaded already by any other deps -> how often is that the case ?
-> Should I revert it from master ?

Maks3w:
Yes this should be revert and keep on the queue until develop is target for next major (v3)

marc-mabe:
I think it's much fine in 2.6.0 as this one is very minor

weierophinney:
@Maks3w adding *development* dependencies can be done at any time; adding new requirements can be done at any minor release (as minor releases signal new features). This falls under the former category, and is safe, as it's only necessary for running unit tests.

Maks3w:
This is removing a "non-dev" dependency.

marc-mabe:
@weierophinney @Maks3w is right in the terms that this change removes a non dev requirement as it moves the optional dependency `zend-serializer` form `require` to `require-dev` (It was already in `suggest`).

History:
I removed this dependency already some time ago zendframework/zf2#2960 but it was reintroduced here zendframework/zf2#6861 (https://github.com/zendframework/zf2/commit/60a0b4c122315fbbc40ba5d900578880a67b7c13)
  -> Please don't ask me why I commended totally different there - I can't remember - I'm only a human)

My opinion `zend-serializer` is only required for the serializer plugin and shoudl therefore be an optional dependency. Moving it from `require` to `require-dev` as it's done in this PR totally makes sense but in fact if someone is using the serializer plugin right now without manually adding it as dependency and if no other dependency has depends on this it ends up in a BC break.

It's already merged to master and develop - so there are 3 possibilities:
 1) leaving it merged as is 
 2) revert it from master only
 3) completely revert it from master and develop and reopen this PR for 3.0

What way we should go?

Maks3w:
1 is bc break
2 is not possible. break sync and you will suffer nightmares
3 it's the correct choice

PR #25, opened 2015-10-25
MongoDd: fixed expiration support and removed duplicated tests

marc-mabe:
with an edge case in `MongoDb::internalGetItem()`

### DIFF
--- a/src/Storage/Adapter/MongoDb.php
+++ b/src/Storage/Adapter/MongoDb.php
@@ -135,7 +135,7 @@ class MongoDb extends AbstractAdapter implements FlushableInterface
             }
 
             if ($result['expires']->sec < time()) {
-                $this->internalRemoveItem($key);
+                $this->internalRemoveItem($normalizedKey);
 
                 return;
             }
--- a/src/Storage/Adapter/MongoDb.php
+++ b/src/Storage/Adapter/MongoDb.php
@@ -136,7 +136,6 @@ class MongoDb extends AbstractAdapter implements FlushableInterface
 
             if ($result['expires']->sec < time()) {
                 $this->internalRemoveItem($normalizedKey);
-
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


Maks3w:
Any possibility of add a test?

marc-mabe:
I have seen a test failure in the develop branch. I'll  check if the test fails also in master else I add it.

marc-mabe:
@Maks3w I found out that the test `MongoDbTest::testCachedItemsExpire` only failed sometimes (see https://travis-ci.org/marc-mabe/zend-cache/jobs/87353080) and noticed that the expiration feature wasn't correctly defined in capabilities. It was `minTtl=0` which means this adapter doesn't support expiration of items. So I fixed the capabilities and also removed all capabilities that are not needed to define because of default.

I also noticed that MongoDB natively supports expiration of items by an index but this index wasn't defined -> so I added this index.
 * MongoDB 3.0 https://docs.mongodb.org/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time
 * MongoDb 2.2 https://docs.mongodb.org/v2.2/tutorial/expire-data/
 * https://github.com/marc-mabe/zend-cache/blob/hotfix/MongoDb_fixed_undef_var_key/src/Storage/Adapter/MongoDbResourceManager.php#L106

-> Because this index will only be defined if the collection wasn't pushed into the resource manager I leave the logic in the adapter intact so it will also work with the index.

At this point I also removed the duplicated tests:
 * `MongoDbTest::testCachedItemsExpire()` already done by `CommonAdapterTest::testGetItemReturnsNullOnExpiredItem()`
 * `MongoDbTest::testFlush()` already done by `CommonAdapterTest::testFlush()`

Maks3w:
I've doubts about what expire strategy is tested then. I think we need two test cases for test when MongDb is who expire or when the defined logic

marc-mabe:
mh I don't know why but if I remove the adapter logic to handle expired items all TTL tests fails even with this index set : https://travis-ci.org/marc-mabe/zend-cache/jobs/88634000

So the logic for handling of expired items needs to be in the adapter even with this index and the logic gets covered with the current tests.

Maks3w:
Ok. Seems the expires index is a new feature and its not related with the original issue.

If it is correct could you split the PR and propose the index feature as enhancement?

marc-mabe:
@Maks3w I removed the line enabling the special index

marc-mabe:
mh caverage for the `Storage/Adapter/MongoDb.php` increased from 82.0 to 82.29 but `APC` related caverage get reported as 0 -> I have no idea why as I didn't change something related to it.

marc-mabe:
merged with 5cda3242d1902528123423fec0643580476aba7f and d2cf8dd81a9a7022de037c568e9e20b1b766ebb6

PR #35, opened 2015-10-30
Benchmarks

marc-mabe:
... like the one in zend-eventmanager

I only created benchḿaks for the `Memory` and for the `Filesystem` storage adapters as these are the only one working from scratch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "zendframework/zend-serializer": "^2.6",
         "zendframework/zend-session": "^2.6.2",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/phpunit": "^4.5"
+        "phpunit/phpunit": "^4.5",
+        "phpbench/phpbench": "^0.10.0"
     },
     "suggest": {
         "zendframework/zend-serializer": "Zend\\Serializer component",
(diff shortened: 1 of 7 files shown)


marc-mabe:
@Ocramius Can we merge this? I know it's not a full suite but it's a start.

Ocramius:
@marc-mabe don't have time to manage merges atm, sorry :-(

composer.json
```diff
@@ -22,7 +22,8 @@
         "zendframework/zend-serializer": "^2.6",
         "zendframework/zend-session": "^2.5",
         "fabpot/php-cs-fixer": "1.7.*",
-        "phpunit/PHPUnit": "~4.0"
+        "phpunit/PHPUnit": "~4.0",
+        "athletic/athletic": "dev-master"
```
Ocramius: note: you may want to aim at phpbench (much better maintenance atm). See https://github.com/zendframework/zend-servicemanager/blob/3c3c3e5d090ae6adb945565529a713bdffd8b330/composer.json#L24

Ocramius: The output standard deviation corrections are worth it: instant benefit and no more re-running the benchmarks a dozen times ;-)

marc-mabe:
@Ocramius I have refactored it to use PHPBench now

PR #36, opened 2015-10-31
fixed AbstractAdapter::internalDecrementItems

marc-mabe:
* was calling `decrementItem` instead of `internalDecrementItem`
 * triggered events for bulk decrement as well as for every items

Noted by benchmarks of #35 

Before:
```
ZendBench\Cache\MemoryStorageAdapterBench
    Method Name                    Iterations   Average Time      Ops/second 
    ----------------------------- ------------ ----------------- -------------
    decrementMissingItemsSingle : [       100] [0.0008270716667] [1,209.08507]
    decrementMissingItemsBulk   : [       100] [0.0009380531311] [1,066.03770]
    decrementExistingItemsSingle: [       100] [0.0008216524124] [1,217.05965]
    decrementExistingItemsBulk  : [       100] [0.0009412384033] [1,062.43009]
```

After:
```
ZendBench\Cache\MemoryStorageAdapterBench
    Method Name                    Iterations   Average Time      Ops/second 
    ----------------------------- ------------ ----------------- -------------
    decrementMissingItemsSingle : [       100] [0.0008160185814] [1,225.46229]
    decrementMissingItemsBulk   : [       100] [0.0001660728455] [6,021.45400]
    decrementExistingItemsSingle: [       100] [0.0008193802834] [1,220.43454]
    decrementExistingItemsBulk  : [       100] [0.0001683640480] [5,939.51032]
```

PS: tests are passing but I need to write a new test catching this

PR #38, opened 2015-11-01
Much better test coverage of AbstractAdapter

marc-mabe:
I don't no why all of these lines were commented out. I think it was from initial writing the component as the tests where written for one of the very very first versions (before 2.0).

I have commented all these tests in and refactored the tests to the current version.

### DIFF
--- a/test/Storage/Adapter/AbstractAdapterTest.php
+++ b/test/Storage/Adapter/AbstractAdapterTest.php
@@ -10,6 +10,9 @@
 namespace ZendTest\Cache\Storage\Adapter;
 
 use Zend\Cache;
+use Zend\Cache\Storage\Adapter\AdapterOptions;
+use Zend\Cache\Storage\Plugin\PluginOptions;
+use Zend\Cache\Storage\Capabilities;
 use Zend\Cache\Exception;
 
 /**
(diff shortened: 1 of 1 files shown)


test/Storage/Adapter/AbstractAdapterTest.php
```diff
@@ -338,23 +348,6 @@ public function testInternalHasItemsCallsInternalHasItem()
         $this->assertEquals(['key1'], $rs);
     }
 
-    public function testGetMetadataCallsInternalGetMetadata()
```
marc-mabe: This gets tested by `testGetMetadatas`

marc-mabe:
The failed test is a real bug and will be fixed in #36

test/Storage/Adapter/AbstractAdapterTest.php
```diff
@@ -416,6 +403,10 @@ public function simpleEventHandlingMethodDefinitions()
 
             ['decrementItem', 'internalDecrementItem', ['k', 1], true],
             ['decrementItems', 'internalDecrementItems', [['k1' => 1, 'k2' => 2]], []],
+
+            ['getCapabilities', 'internalGetCapabilities', [], new Capabilities(
```
Maks3w: I prefer create the object in a local var with a descriptive name so you can read the list of test cases quickly

Maks3w:
I'll review this later

test/Storage/Adapter/AbstractAdapterTest.php
marc-mabe: updated

PR #40, opened 2015-11-02
hotfix : Redis.addItem TTL

jbottigliero:
- Updates the Redis adapter's `internalAddItem` method to utilize TTL configuration.

I do have a Redis adapter specific test ready to be merged in, [here](https://github.com/jbottigliero/zend-cache/pull/1/files). 

However, I wasn't sure if it's something that should actually be placed into the `CommonAdapterTest` as [other](https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/Apc.php#L456
) [adapters](https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/Memcache.php#L378) are utilizing the functionality, but it's not explicitly being tested (`ttl` via `addItem` specifically) as far as I can tell.

### DIFF
--- a/test/Storage/Adapter/CommonAdapterTest.php
+++ b/test/Storage/Adapter/CommonAdapterTest.php
@@ -720,6 +720,32 @@ abstract class CommonAdapterTest extends \PHPUnit_Framework_TestCase
         $this->assertTrue($this->_storage->hasItem('key2'));
     }
 
+    public function testAddItemSetsTTL()
+    {
+        $capabilities = $this->_storage->getCapabilities();
+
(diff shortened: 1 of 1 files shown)


marc-mabe:
Thanks for your PR.

Yes please add a test case in `CommonAdapterTest` - as all adapter supporting that feature should work the same or differences should be notable by `Capabilities`.

Your linked test looks fine :+1:

jbottigliero:
Updated PR to include `CommonAdapterTest.testAddItemSetsTTL` – let me know if this isn't what you had in mind!

marc-mabe:
merged thanks!

Issue #47, opened 2015-11-12
Discussion: APC vs APCU


remicollet:
About Zend\Cache\Storage\Adapter\Apc

Starting with PHP 7, APCU won't provide the apc_* fonction by default. So the test in __construct will pass, and later other methods will fail with undefined function.

For "legacy" branch

     $enabled = ini_get('apc.enabled') && function_exists('apc_store');

For current branch which requires PHP >= 5.5, as no APC version exists I think It could make sense to use APCU by default, and switch all call to apcu_* functions.

Another way could be to create an Zend\Cache\Storage\Adapter\Apcu adapter, and mark the previous as deprecated. But it will require code changes in all consumer of this library.

Comments ?

P.S. notice, for now, apcu 5.0.0-dev segfaults during the test suite.... under investigation.

Maks3w:
If no BC break exists I think we can modify APC class for make internal calls to APCU.

At a second step we could rename the class with the correct "APCU" name and deprecate the old name. However there is no real wins of this so this could be do later.

remicollet:
> If no BC break exists I think we can modify APC class for make internal calls to APCU.

I have to code ready on my computer, I'm waiting to ba able to run the test suite, before submitting a PR
https://github.com/remicollet/zend-cache/commits/issue-apcu

marc-mabe:
@remicollet @Maks3w I would prefer to add an Apcu adapter and make the Apc adapter as deprecated as I noticed that HHVM only supports the old APC methods.

I already worked on an Apcu adapter but currently I get some strange failures on travis that I don't get on my local machine.
-> https://github.com/marc-mabe/zend-cache/tree/feature/23/src/Storage/Adapter
-> https://travis-ci.org/marc-mabe/zend-cache/builds/90415153

marc-mabe:
Link to the other discussion : #23

remicollet:
> I would prefer to add an Apcu adapter 

I understand, even if I will prefer something transparent for users...

marc-mabe:
To summarize as I understand:

* The Apc adapter was originally developed to be a wrapper for the APC extension
* Later it was going to be also a wrapper for others that support the same interface as the APC extension (APCu and HHVM)
* HHVM has (or had) a small difference in it metadata reporting - see https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/Apc.php#L717
* APCu fully supports the APC interface but only if compiled with APC compatibility 
[...]

PR #48, opened 2015-11-27
New adapter for APCU 5.1 (PHP 7)

remicollet:
See #47

### DIFF
--- a/src/Storage/Adapter/ApcuOptions.php
+++ b/src/Storage/Adapter/ApcuOptions.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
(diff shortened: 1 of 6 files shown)


src/Storage/Adapter/ApcuOptions.php
```diff
@@ -0,0 +1,47 @@
...
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Cache\Storage\Adapter;
+
+/**
+ * These are options specific to the APC adapter
+ */
+class ApcuOptions extends AdapterOptions
```
Maks3w: This class could be removed and reuse ApcOptions

Maks3w:
Please extend from APC classes and override the methods where needed.

remicollet:
Please consider this mostly as a PoC.

I'm not enough aware of ZF internals to create something perfect, and don't want to spend my time bikesheeding.

At least, now, I know that it "can" work (with latest APCU 5.1 without apc compat ext), while it is for now broken.

Feel free to rewrite this as you prefer.

Maks3w:
Thanks for the PoC. Unfortunately this can't be merged as is.

Issue #23, opened 2015-10-25
Add APCu adapter as feature replacement for Apc


DCMNMarc:
As of the old PHP extension [APC](http://pecl.php.net/package/APC) seems inactive and doesn't compile with PHP-5.6 / PHP-7 (PHP-5.4 should work but PHP-5.5 not tested) we can drop the adapter for it.

In fact the APC extension is inactive because of the PHP opcache and the possibility to store user data in the cache has been moved to [APCu](http://pecl.php.net/package/APCu) long time ago.

APC and APCu is nearly compatible on storing user data but in some cases it's not. The differences are already part of the Apc adapter.

To simplify the adapter a little bit and because it's no longer possible to use the old APC extension with the minimal required PHP version it makes sense to completely drop support for it.

Renaming the adapter from Apc to Apcu:
If I only take a look what the optimal way would look like it would make sense to also rename the Apc adapter to Apcu. With a deeper look the there are two cases where it could make confusion.
* The APCu extension comes with a BC safe version (the version that the current adapter supports only)
 * Even if it's compiled with BC to APC all APCu functionalities are available
 * -> from this POV it's fine
* HHVM comes with APC support for the old APC extension.
 * Does someone now if HHVM also supports APCu?

@weierophinney @ezimuel @Ocramius @Maks3w Thoughts?

Ocramius:
The point of APCu is to make the switch to it transparent: do not change
the APC adapter unless there is an actual need for it, and the current
adapter works just fine with APCu.

marc-mabe:
@Ocramius  The current adapter only works with APCu if compiled with the flag `--enable-apc-bc`

Ocramius:
@marc-mabe which is the only reason why somebody would use `APCu` anyway? :|

marc-mabe:
@Ocramius The APC extension is dead and it doesn't work with our out min requirements. The most common way to cache user data into shared memory is APCu but we only support APCu if it was compiled with the BC flag. I think it would be very good if we completely support APCu - not only in one outdated case.

Btw. the default option for this flag is to not compile it with for BC ;)

Ocramius:
> Btw. the default option for this flag is to not compile it with for BC ;)

Yes, but no APCu installation that I saw so far uses APCu specific stuff: everyone just wants the BC compliant interface.

 > I think it would be very good if we completely support APCu - not only in one outdated case.

[...]

PR #51, opened 2015-11-29
Apcu adapter as future replacement for Apc

marc-mabe:
link #23 #47 #48 

It adds an additional storage adapter for APCu to work with native `apcu_`  functions instead of the combat `apc_` functions.

It requires APCu 5.1.0 as this fixes the compatibility issue with `APCIterator` by introducing a new `APCuIterator`. Older versions should be used in combat mode together with the old Apc adapter.

It doesn't extend the old adapter because:
* It would introduce a future BC break if removing the old APC adapter
* It would require to overwrite ~80% of the functions only to replace `apc_` with `apcu_` or would require still a lot refactoring

PS: Travis is running only this new adapter for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ services:
 env:
   global:
     - TESTS_ZEND_CACHE_APC_ENABLED: true
+    - TESTS_ZEND_CACHE_APCU_ENABLED=true
     - TESTS_ZEND_CACHE_MEMCACHED_ENABLED: true
     - TESTS_ZEND_CACHE_MEMCACHED_HOST: "127.0.0.1"
     - TESTS_ZEND_CACHE_MEMCACHED_PORT: 11211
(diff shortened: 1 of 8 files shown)


marc-mabe:
@Maks3w It's ready now - could you have a look
PS: It's currently marked as `3.0` but it could also `2.next` even if the adapter requires a very new extension version and the extension itself requires PHP-7.

Maks3w:
Why duplicate every class like #48 did?

marc-mabe:
@Maks3w even if we extend one adapter over the other would still require to overwrite 90% of the functions to differ between `apc_*`  and `apcu_*` function calls or totally refactor the adapter into something like we did with the `AbstractZendServer` adapters and if we go to deprecate the old `Apc` adapter we would introduce a BC break. Additionally the `Apc` adapter has some extra logic to make it compatible with the previous `apcu` extension in BC mode and the one of HHVM.

marc-mabe:
@Maks3w Anything left to say?

Maks3w:
No, you know more about this component than me. I just can notice if something in the design is weird for me.

marc-mabe:
Ok so I can merge this PR

marc-mabe:
rebased against develop

PR #52, opened 2015-12-21
fixed #50: run APC(u) tests in travis

### LINKED ISSUE #50: Travis: APC tests not running
I found out two reasons:

1) Installing the APCu extension fails on PHP < 7 as the last beta supports PHP 7 only
   `echo "yes\nno\n" | pecl -d preferred_state=beta install apcu || return 0`
2) It looks like the extension will not be automatically enabled by `pecl`

marc-mabe:
This (re)-enables APC adapter tests on travis (#50)
- using apcu-4.0.8 (currently last stable) extension on PHP < 7
- using apcu_bc (last beta) extension on PHP 7
  - There is a bug using `pecl install apcu_bc-beta` where it attaches the directive `extension=apc.so` on top of php.ini which results in wrong extension loading :(
  - To solve that issue I added another ini file loading the extension in bottom which still results in a warning of the first load but the test works
  -> @Maks3w other ideas ?

### DIFF
--- a/.ci/apcu_bc.ini
+++ b/.ci/apcu_bc.ini
@@ -1,0 +1,1 @@
+extension=apc.so
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,16 @@ matrix:
     - php: 5.5
       env:
         - EXECUTE_CS_CHECK=true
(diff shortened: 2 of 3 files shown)


Maks3w:
ok for me. Workaround is enough documented.

PR #53, opened 2015-12-21
fixed seg fault in redis adapter on PHP 7

marc-mabe:
After changing selected database on an already connected object
the `RedisResourceManager` was again trying to connect which crashed.

### DIFF
--- a/src/Storage/Adapter/RedisResourceManager.php
+++ b/src/Storage/Adapter/RedisResourceManager.php
@@ -332,7 +332,7 @@ class RedisResourceManager
             $resource = array_merge(
                 $defaults,
                 [
-                    'resource' => $resource,
+                    'resource'    => $resource,
                     'initialized' => isset($resource->socket),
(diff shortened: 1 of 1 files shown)


marc-mabe:
To test added as the current tests reflected this seg fault already on PHP 7

gianarb:
Sorry little question,

is a good practice self merge own PRs without any review?

For me this PR is perfect :) But I don't know if its the correct flow.

marc-mabe:
hi @gianarb, normally you are right! I did it only this time because it's a simple fix and it breaks very hard. It also breaks nearly all tests running on PHP 7 where it makes it very hard to find out if other changes are correct.

Please tell me I should avoid this practice in the future and/or what I can do if it doesn't get reviewed for a long time (which is sometimes the case here as this component requires a lot of setup to run).

gianarb:
No problem for me :) 
Thanks for your work!

Issue #46, opened 2015-11-11
PSR-6 wrapper?


kynx:
Has anyone done any work on a PSR-6 wrapper for zend-cache? I know the PSR still in review, but it would be darn handy...

weierophinney:
Considering there are still a few proposed changes to the spec, we can't
incorporate such a wrapper in a stable version. Once the spec is accepted,
absolutely we'd like to do this.

kynx:
Thanks for that. Yes, I appreciate it's not going to land in stable until accepted - was more wondering if anyone had a prototype they'd been working on. If not, I'll just have to see how itchy I get :wink:

kynx:
Well, the itching got the better of me. My first stab at this is [over here](https://github.com/kynx/zend-cache-psr6) if anyone's interested. Still a [bunch of things](https://github.com/kynx/zend-cache-psr6/blob/master/TODO.md) to iron out, but more than happy to contribute this back when (surely not if?) PSR-6 finally gets accepted.

kynx:
PSR-6 finally got accepted! 

There's some useful stuff in FIG's [cache-util](https://github.com/php-fig/cache-util) that I need to work into [my wrapper](https://github.com/kynx/zend-cache-psr6), and some thought to put into expiry / clearing / flushing. But if no one else has plans in progress I'll be happy to take a break from the mince pies over the next week or so :smile:

marc-mabe:
@kynx #80 has been merged into develop now. Please can you update your PR and not allow adapters capability `lockOnExpire > 0` thanks!

marc-mabe:
closed by cf35782

PR #54, opened 2016-01-03
Added PSR-6 adapter

kynx:
This PR address #46, adding a [PSR-6](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md) wrapper.

Some key points:

* Not all storage adapters are supported, due the PSR-6 requirement that individual cache items must support separate TTLs, set on save. This rules out Dba, Filesystem, Memory, Session and Redis < v2
* PSR-6 [Error Handling](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md#error-handling) requires that most exceptions thrown by the storage adapter be suppressed. I've provided an `ExceptionLogger` so these can be saved to a PSR-3 compatible logger. Some documentation is probably in order!
* There are some useful integration tests at https://github.com/php-cache/integration-tests. I haven't got around to incorporating those yet - only discovered them yesterday.

Happy New Year!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,11 +16,13 @@
         "php": "^5.5 || ^7.0",
         "zendframework/zend-stdlib": "^2.7 || ^3.0",
         "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
-        "zendframework/zend-eventmanager": "^2.6.2 || ^3.0"
+        "zendframework/zend-eventmanager": "^2.6.2 || ^3.0",
+        "psr/cache": "^1.0"
     },
     "require-dev": {
         "zendframework/zend-serializer": "^2.6",
         "zendframework/zend-session": "^2.5",
+        "cache/integration-tests": "^0.7.0",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require-dev": {
         "zendframework/zend-serializer": "^2.6",
         "zendframework/zend-session": "^2.5",
-        "cache/integration-tests": "^0.7.0",
+        "cache/integration-tests": "^0.8.0",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "~4.0"
     },
--- a/doc/book/psr6.md
+++ b/doc/book/psr6.md
@@ -51,7 +51,9 @@ found.
 ## Supported adapters
 
 The PSR-6 specification requires that the underlying storage support time-to-live (TTL), which is set when the
-item is saved. For this reason the following adapters cannot be used: `Dba`, `Filesystem`, `Memory` and `Session`.
+item is saved. For this reason the following adapters cannot be used: `Dba`, `Filesystem`, `Memory` and `Session`. The 
+`XCache` adapter calculates TTLs based on the request time, not the time the item is actually persisted, which means 
+that it also cannot be used.
 
 In addition adapters must support the `Zend\Cache\FlushableInterface`. All the current `Zend\Cache\Storage\Adapter`s 
 fulfil this requirement.
--- a/src/Psr/CacheItem.php
+++ b/src/Psr/CacheItem.php
@@ -6,6 +6,7 @@
  * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
+
 namespace Zend\Cache\Psr;
 
 use DateInterval;
--- a/src/Psr/CacheItem.php
+++ b/src/Psr/CacheItem.php
@@ -1,0 +1,173 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/).
+ *
+ * @link      http://github.com/zendframework/zend-cache for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+namespace Zend\Cache\Psr;
+
+use DateInterval;
+use DateTime;
+use DateTimeInterface;
+use DateTimeZone;
+use Psr\Cache\CacheItemInterface;
+
+final class CacheItem implements CacheItemInterface
+{
+    /**
+     * Cache key
+     * @var string
+     */
+    private $key;
+
+    /**
+     * Cache value
+     * @var mixed|null
+     */
+    private $value;
+
+    /**
+     * True if the cache item lookup resulted in a cache hit or if they item is deferred or successfully saved
+     * @var bool
+     */
+    private $isHit = false;
+
+    /**
+     * Timestamp item will expire at if expiresAt() called, null otherwise
+     * @var int|null
+     */
+    private $expiration = null;
+
+    /**
+     * Seconds after item is stored it will expire at if expiresAfter() called, null otherwise
+     * @var int|null
+     */
+    private $ttl = null;
+
+    /**
+     * @var DateTimeZone
+     */
+    private $tz;
+
+    /**
+     * Constructor.
+     *
+     * @param string $key
+     * @param mixed $value
+     * @param bool $isHit
+     */
+    public function __construct($key, $value, $isHit)
+    {
+        $this->key = $key;
+        $this->value = $isHit ? $value : null;
+        $this->isHit = $isHit;
+        $this->tz = new DateTimeZone('UTC');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function get()
+    {
+        return $this->value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isHit()
+    {
+        if (! $this->isHit) {
+            return false;
+        }
+        $ttl = $this->getTtl();
+        return $ttl === null || $ttl > 0;
+    }
+
+    /**
+     * Sets isHit value
+     *
+     * This function is called by CacheItemPoolAdapter::saveDeferred() and is not intended for use by other calling
+     * code.
+     *
+     * @param boolean $isHit
+     * @return $this
+     */
+    public function setIsHit($isHit)
+    {
+        $this->isHit = $isHit;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function set($value)
+    {
+        $this->value = $value;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAt($expiration)
+    {
+        if (! ($expiration === null || $expiration instanceof DateTimeInterface)) {
+            throw new InvalidArgumentException('$expiration must be null or an instance of DateTimeInterface');
+        }
+
+        $this->expiration = $expiration instanceof DateTimeInterface ? $expiration->getTimestamp() : null;
+        $this->ttl = null;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAfter($time)
+    {
+        if ($time instanceof DateInterval) {
+            $end = new DateTime('now', $this->tz);
+            $end->add($time);
+            $this->ttl = $end->getTimestamp() - time();
+        } elseif (is_int($time) || $time === null) {
+            $this->ttl = $time;
+        } else {
+            throw new InvalidArgumentException(sprintf('Invalid $time "%s"', gettype($time)));
+        }
+
+        $this->expiration = null;
+
+        return $this;
+    }
+
+    /**
+     * Returns number of seconds until item expires
+     *
+     * If NULL, the pool should use the default TTL for the storage adapter. If <= 0, the item has expired.
+     *
+     * @return int|null
+     */
+    public function getTtl()
+    {
+        $ttl = $this->ttl;
+        if ($this->expiration !== null) {
+            $ttl = $this->expiration - time();
+        }
+        return $ttl;
+    }
+}
--- a/src/Psr/CacheItemPoolAdapter.php
+++ b/src/Psr/CacheItemPoolAdapter.php
@@ -11,7 +11,6 @@ namespace Zend\Cache\Psr;
 use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Zend\Cache\Exception;
-use Zend\Cache\Storage\Adapter\Apc;
 use Zend\Cache\Storage\ClearByNamespaceInterface;
 use Zend\Cache\Storage\FlushableInterface;
 use Zend\Cache\Storage\StorageInterface;
@@ -321,18 +320,18 @@ class CacheItemPoolAdapter implements CacheItemPoolInterface
             ));
         }
 
-        if ($storage instanceof Apc && ini_get('apc.use_request_time')) {
-            throw new CacheException(sprintf(
-                'Cannot use %s with apc.use_request_time = 1',
-                get_class($storage)
-            ));
-        }
-
         // we've got to be able to set per-item TTL on write
         $capabilities = $storage->getCapabilities();
         if (!($capabilities->getStaticTtl() && $capabilities->getMinTtl())) {
             throw new CacheException(sprintf(
                 'Storage %s does not support static TTL',
+                get_class($storage)
+            ));
+        }
+
+        if ($capabilities->getUseRequestTime()) {
+            throw new CacheException(sprintf(
+                'Cannot use %s with useRequestTime = true',
                 get_class($storage)
             ));
         }
--- a/src/Psr/CacheItemPoolAdapter.php
+++ b/src/Psr/CacheItemPoolAdapter.php
@@ -1,0 +1,420 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/).
+ *
+ * @link      http://github.com/zendframework/zend-cache for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+namespace Zend\Cache\Psr;
+
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Zend\Cache\Exception;
+use Zend\Cache\Storage\Adapter\Apc;
+use Zend\Cache\Storage\ClearByNamespaceInterface;
+use Zend\Cache\Storage\FlushableInterface;
+use Zend\Cache\Storage\StorageInterface;
+
+/**
+ * PSR-6 cache adapter
+ *
+ * @link https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md
+ * @link https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache-meta.md
+ */
+class CacheItemPoolAdapter implements CacheItemPoolInterface
+{
+    /**
+     * @var StorageInterface
+     */
+    private $storage;
+
+    /**
+     * @var CacheItem[]
+     */
+    private $deferred = [];
+
+    /**
+     * @var bool
+     */
+    private $serializeValues = false;
+
+    /**
+     * @var string
+     */
+    private static $serializedFalse;
+
+    /**
+     * Constructor.
+     *
+     * PSR-6 requires that all implementing libraries support TTL so the given storage adapter must also support static
+     * TTL or an exception will be raised. Currently the following adapters do *not* support static TTL: Dba,
+     * Filesystem, Memory, Session and Redis < v2.
+     *
+     * @param StorageInterface $storage
+     *
+     * @throws CacheException
+     */
+    public function __construct(StorageInterface $storage)
+    {
+        $this->validateStorage($storage);
+
+        $this->serializeValues = $this->shouldSerialize($storage);
+        if ($this->serializeValues) {
+            static::$serializedFalse = serialize(false);
+        }
+
+        $this->storage = $storage;
+    }
+
+    /**
+     * Destructor.
+     *
+     * Saves any deferred items that have not been committed
+     */
+    public function __destruct()
+    {
+        $this->commit();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItem($key)
+    {
+        $this->validateKey($key);
+
+        if (! $this->hasDeferredItem($key)) {
+            $value = null;
+            $isHit = false;
+            try {
+                $value = $this->storage->getItem($key, $isHit);
+
+                if ($this->serializeValues && $isHit) {
+                    // will set $isHit = false if unserialization fails
+                    extract($this->unserialize($value));
+                }
+            } catch (Exception\InvalidArgumentException $e) {
+                throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+            } catch (Exception\ExceptionInterface $e) {
+                // ignore
+            }
+
+            return new CacheItem($key, $value, $isHit);
+        } else {
+            return clone $this->deferred[$key];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItems(array $keys = [])
+    {
+        $this->validateKeys($keys);
+        $items = [];
+
+        // check deferred items first
+        foreach ($keys as $key) {
+            if ($this->hasDeferredItem($key)) {
+                // dereference deferred items
+                $items[$key] = clone $this->deferred[$key];
+            }
+        }
+
+        $keys = array_diff($keys, array_keys($items));
+
+        if (count($keys)) {
+            try {
+                $cacheItems = $this->storage->getItems($keys);
+            } catch (Exception\InvalidArgumentException $e) {
+                throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+            } catch (Exception\ExceptionInterface $e) {
+                $cacheItems = [];
+            }
+
+            foreach ($cacheItems as $key => $value) {
+                $isHit = true;
+                if ($this->serializeValues) {
+                    // will set $isHit = false if unserialization fails
+                    extract($this->unserialize($value));
+                }
+
+                $items[$key] = new CacheItem($key, $value, $isHit);
+            }
+
+            // Return empty items for any keys that where not found
+            foreach (array_diff($keys, array_keys($cacheItems)) as $key) {
+                $items[$key] = new CacheItem($key, null, false);
+            }
+        }
+
+        return $items;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasItem($key)
+    {
+        $this->validateKey($key);
+
+        // check deferred items first
+        $hasItem = $this->hasDeferredItem($key);
+
+        if (!$hasItem) {
+            try {
+                $hasItem = $this->storage->hasItem($key);
+            } catch (Exception\InvalidArgumentException $e) {
+                throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+            } catch (Exception\ExceptionInterface $e) {
+                $hasItem = false;
+            }
+        }
+
+        return $hasItem;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * If the storage adapter supports namespaces and one has been set, only that namespace is cleared; otherwise
+     * entire cache is flushed.
+     */
+    public function clear()
+    {
+        $this->deferred = [];
+
+        try {
+            $namespace = $this->storage->getOptions()->getNamespace();
+            if ($this->storage instanceof ClearByNamespaceInterface && $namespace) {
+                $cleared = $this->storage->clearByNamespace($namespace);
+            } else {
+                $cleared = $this->storage->flush();
+            }
+        } catch (Exception\ExceptionInterface $e) {
+            $cleared = false;
+        }
+
+        return $cleared;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItem($key)
+    {
+        return $this->deleteItems([$key]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItems(array $keys)
+    {
+        $this->validateKeys($keys);
+
+        $deleted = true;
+
+        // remove deferred items first
+        $this->deferred = array_diff_key($this->deferred, array_flip($keys));
+
+        try {
+            $this->storage->removeItems($keys);
+        } catch (Exception\InvalidArgumentException $e) {
+            throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+        } catch (Exception\ExceptionInterface $e) {
+            $deleted = false;
+        }
+
+        return $deleted;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function save(CacheItemInterface $item)
+    {
+        if (!$item instanceof CacheItem) {
+            throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
+        }
+
+        $saved = true;
+        $options = $this->storage->getOptions();
+        $ttl = $options->getTtl();
+        try {
+            // serialize value, if required
+            $value = $item->get();
+            if ($this->serializeValues) {
+                $value = serialize($value);
+            }
+
+            // reset TTL on adapter, if required
+            $itemTtl = $item->getTtl();
+            if ($itemTtl > 0) {
+                $options->setTtl($itemTtl);
+            }
+
+            // don't save items that have already expired
+            if ($itemTtl === null || $itemTtl > 0) {
+                $saved = $this->storage->setItem($item->getKey(), $value);
+                // saved items are a hit? see integration test CachePoolTest::testIsHit()
+                $item->setIsHit($saved);
+            }
+        } catch (Exception\InvalidArgumentException $e) {
+            throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+        } catch (Exception\ExceptionInterface $e) {
+            $saved = false;
+        } finally {
+            $options->setTtl($ttl);
+        }
+
+        return $saved;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function saveDeferred(CacheItemInterface $item)
+    {
+        if (!$item instanceof CacheItem) {
+            throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
+        }
+
+        // deferred items should always be a 'hit' until they expire
+        $item->setIsHit(true);
+        $this->deferred[$item->getKey()] = $item;
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function commit()
+    {
+        $notSaved = [];
+
+        foreach ($this->deferred as &$item) {
+            if (! $this->save($item)) {
+                $notSaved[] = $item;
+            }
+        }
+        $this->deferred = $notSaved;
+
+        return empty($this->deferred);
+    }
+
+    /**
+     * Throws exception is storage is not compatible with PSR-6
+     * @param StorageInterface $storage
+     * @throws CacheException
+     */
+    private function validateStorage(StorageInterface $storage)
+    {
+        // all current adapters implement this
+        if (!$storage instanceof FlushableInterface) {
+            throw new CacheException(sprintf(
+                'Storage %s does not implement %s',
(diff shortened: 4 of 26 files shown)


Nyholm:
The implementation looks good. There were some small bugs thou.. Say you do like this: 
```php
$item = $pool->getItem('key');
$item->set('foo');
$pool->saveDeferred($item);

$item = $pool->getItem('key');
$item->set('bar');
// No save 

$item = $pool->getItem('key');
$item->get(); // should be 'foo'

```



I added some integration tests to this PR: https://github.com/kynx/zend-cache/pull/1

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        return $this->key;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function get()
+    {
+        return $this->value;
```
Nyholm: You need to check if isHit before you return anything

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isHit()
+    {
+        return $this->isHit;
```
Nyholm: Say that you pull something from cache, waits 10 seconds and then the cache expires. Then isHit() should be false. 

This function needs to verity the expiration time

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function save(CacheItemInterface $item)
+    {
+        if (!$item instanceof CacheItem) {
+            throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
+        }
+
+        $this->validateKey($item->getKey());
```
Nyholm: No *need* to validate the key. Calling libraries should never instantiate a `CacheItem`. So the key is already checked by `getItems`.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+     * {@inheritdoc}
+     */
+    public function saveDeferred(CacheItemInterface $item)
+    {
+        if (!$item instanceof CacheItem) {
+            throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
+        }
+
+        $key = $item->getKey();
+        $this->validateKey($key);
```
Nyholm: No *need* to validate the key. Se comment above.

src/Psr/CacheItem.php
kynx: Are you 100% sure about this? The docblock states "Confirms if the cache item lookup resulted in a cache hit", and I can't find anything in the rest of the spec that specifically talks about expiration post-lookup.

Doing it any other way with Zend Cache would not be reliable: TTL is not accessible on fetch per-item - we only have the default TTL for the storage adapter.

src/Psr/CacheItem.php
kynx: The value is set to `null` in the constructor if the item isn't a hit.

src/Psr/CacheItem.php
Nyholm: Sorry, you are correct. I miss-remembered this paragraph: 

> Hit - A cache hit occurs when a Calling Library requests an Item by key and a matching value is found for that key, and that value has not expired, and the value is not invalid for some other reason. Calling Libraries SHOULD make sure to verify isHit() on all get() calls.

It says "Calling Libraries SHOULD make sure to verify isHit() on all get() calls.". I remembered it as  "Implementing Libraries... ". 

So a there is not need to change `$this->isHit = false` at any time in your implementation.

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+        if ($time instanceof DateInterval) {
+            $interval = $time;
+        } elseif ((int) $time == $time) {
+            $interval = new DateInterval('PT' . $time . 'S');
+        } else {
+            throw new InvalidArgumentException(sprintf('Invalid $time "%s"', gettype($time)));
+        }
+
+        $this->expiration = new \DateTime();
+        $this->expiration->add($interval);
```
marc-mabe: I'm not sure if this is safe for all time zones as in dst time zones there are times ambiguous.
see https://3v4l.org/BnPd0 (check the diff)
-> Can we force the time zone as UTC ?

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        if ($time instanceof DateInterval) {
+            $interval = $time;
+        } elseif ((int) $time == $time) {
+            $interval = new DateInterval('PT' . $time . 'S');
+        } else {
+            throw new InvalidArgumentException(sprintf('Invalid $time "%s"', gettype($time)));
+        }
+
+        $this->expiration = new \DateTime();
```
marc-mabe: `use DateTime;`

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAfter($time)
+    {
+        if ($time instanceof DateInterval) {
+            $interval = $time;
+        } elseif ((int) $time == $time) {
```
marc-mabe: Shouldn't that be `is_int($time)`?

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,130 @@
...
+            $interval = new DateInterval('PT' . $time . 'S');
+        } else {
```
marc-mabe: I'm missing the handling of `null` to get the default TTL defined by adapter

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+            throw new CacheException(sprintf(
+                'Storage %s does not implement %s',
+                get_class($storage),
+                FlushableInterface::class
+            ));
+        }
+
+        // we've got to be able to set per-item TTL on write
+        $capabilities = $storage->getCapabilities();
+        if (!($capabilities->getStaticTtl() && $capabilities->getMinTtl())) {
```
marc-mabe: I think the check for `&& $capabilities->getMinTtl()` should be removed as this only describes that it's possible to store an item without expiration but not that the adapter doesn't support expiration.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+    public function clear()
+    {
+        $this->deferred = [];
+
+        try {
+            $namespace = $this->storage->getOptions()->getNamespace();
+            if ($this->storage instanceof ClearByNamespaceInterface && $namespace) {
+                $cleared = $this->storage->clearByNamespace($namespace);
+            } else {
+                $cleared = $this->storage->flush();
```
marc-mabe: Is there any description about that difference in PRS-6?

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItem($key)
+    {
+        $items = $this->getItems([$key]);
+
+        return array_pop($items);
+    }
```
marc-mabe: As single items are used much more in user land than a bunch of items I prefer to remove the small overhead of wrapping a single item call into a bunch of one item.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,287 @@
...
+            }
+        }
+        $this->deferred = $notSaved;
+
+        return empty($this->deferred);
+    }
+
+    private function validateKey($key)
+    {
+        if (!is_string($key) || preg_match('#[{}()/\\\\@:]#', $key)) {
```
marc-mabe: We should find a way to validate the key by adapter as some adapters have additional limitations.

marc-mabe:
Hi @kynx thanks for your work!

I have added some notes from a quick look.
What I don't understand is why did you addad the PSR-3 `ExceptionLogger`? In my opinion this has nothing to do with the PSR-6 implementation and we already have a storage adapter plugin you know where you can handle all thrown exceptions as you like.

src/Psr/CacheItemPoolAdapter.php
kynx: I'm catching and re-throwing any `InvalidArgumentException`s thrown by the storage adapter to cover this. It's a bit blunt - they could be thrown outside `AbstractAdapter::normalizeKey()`, though I couldn't spot this in the code paths I'm using. But if that's a worry it wouldn't be hard to cut-and-paste the `normalizeKey()` logic.

src/Psr/CacheItemPoolAdapter.php
kynx: No, the spec has no concept of namespaces. I'm doing this to avoid clobbering cache items outside the namespace set up for the storage adapter, if it supports them.

src/Psr/CacheItemPoolAdapter.php
kynx: Heh. I actually nicked that from your code about 2/3 of the way down https://github.com/zendframework/zf2/pull/5386 :smile: 

Redis actually sets `staticTtl = true` and `minTtl = 0` if version < 2. Not entirely sure what that means, but decided to play safe.

marc-mabe: Oh wow you are totally right - I shouldn't comment with my head somewhere else.
-> Please ignore my previous comment

src/Psr/CacheItemPoolAdapter.php
marc-mabe: mh PSR-6 has no namespace definition but the definition of a pool could be interpreted for us as a collection of items of one namespace of the same storage.
From this POV it would be OK, even to flush the hole cache storage as fallback would effect items of another pool but for a cache and the spec there is no guarantee that items exists as long as they are not deleted or expired. So it only effects performance of the different pool.

:+1:

kynx:
@marc-mabe Many thanks for the feedback.

The `ExceptionLogger` is there for a couple of reasons:

1. The [spec](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md#error-handling) says: "An Implementing Library SHOULD log such errors or otherwise report them to an administrator as appropriate."
2. There's a chance a user will attach an `ExceptionHandler` plugin that has `'throw_exceptions' => false`. That would screw up the boolean return values of most methods that talk to storage if it failed. So the `ExceptionLogger` will barf if it finds another `ExceptionHandler` plugin attached.

I didn't want to completely rule out users handling exceptions however they wanted, but I did imagine that any Service class would add the logger by default - even if it defaulted to a `NullLogger`. But I'll ditch it if you want.

Let me know if there's more to do with `validateKey()` and the `clear()` stuff.

kynx:
@Nyholm Many thanks for the feedback and integration tests. I've mocked out the adapter in those now and made sure they all pass :smiley:

Nyholm:
> @Nyholm Many thanks for the feedback and integration tests. I've mocked out the adapter in those now and made sure they all pass :smiley:

You are welcome. 

I see that your mock preforms better than an actual adapter. When I test this with the redis adapter I fail a few tests. The redis adapter does not preserve datatypes. Consider this test:

```php
$item = $this->cache->getItem('key');
$item->set(5); // Integer 5
$this->cache->save($item);

$item = $this->cache->getItem('key');
$this->assertTrue(5 === $item->get(), 'Wrong data type. If we store an int we must get an int back.');
```

Also, this just crossed my mind. This is not a part of PSR-6 as far as I know. But what if you do like this: 
```php
$item = $this->cache->getItem('key');
$item->set('foo');
$item->expiresAfter(3600);
$this->cache->save($item);

// 10 minutes later in an other request
$item = $this->cache->getItem('key');
$item->set('bar');
$this->cache->save($item);
```
I'm not too sure how the storage options work. For how long will the item be saved? 3000 more seconds? 3600 seconds? Or infinity?

kynx:
Yeah, I noticed that fail with Redis too. Do you happen to know if this is a general issue with how Redis serializes data, or if it's specific to Zend Cache? 

OT, but it would be useful to provide a mechanism in the integration tests for marking certain tests skipped - maybe be just an associative array of methodName => reason that gets checked at the start of each test.

Regarding re-saving an item after TTL has been set: the default TTL configured for the storage would be used in this case. I don't think any implementation could do otherwise reliably - another process may have cleared the cache, it might be unavailable, etc.

marc-mabe:
> The ExceptionLogger is there for a couple of reasons:
>
>    The spec says: "An Implementing Library SHOULD log such errors or otherwise report them to an administrator as appropriate."
>    There's a chance a user will attach an ExceptionHandler plugin that has 'throw_exceptions' => false. That would screw up the boolean return values of most methods that talk to storage if it failed. So the ExceptionLogger will barf if it finds another ExceptionHandler plugin attached.

The spec makes a recommendation to log errors or report them but nobody as the calling library can actually handle errors. Also I don't see a difference where your `ExceptionLogger` makes it in any case simpler to log errors. `zend-cache` already makes it very simple to handle errors.

> Yeah, I noticed that fail with Redis too. Do you happen to know if this is a general issue with how Redis serializes data, or if it's specific to Zend Cache? 

Redis has no serialize feature, it simply gets requires a string and PHP converts any given value into a string at this point.

From spec
```
Implementing libraries MUST support all serializable PHP data types, including:

    Strings - Character strings of arbitrary size in any PHP-compatible encoding.
    Integers - All integers of any size supported by PHP, up to 64-bit signed.
    Floats - All signed floating point values.
    Boolean - True and False.
    Null - The actual null value.
    Arrays - Indexed, associative and multidimensional arrays of arbitrary depth.
    Object - Any object that supports lossless serialization and deserialization such that $o == unserialize(serialize($o)). Objects MAY leverage PHP's Serializable interface, __sleep() or __wakeup() magic methods, or similar language functionality if appropriate.
```

That means you have to make sure the adapter (or the serializer plugin) serializes data (reported by `getCapabilities()->getSupportedDataTypes()`) and if not you need to `un/serialize` the data by hand to match the spec.

kynx:
I've dropped `ExceptionLogger`.

My holidays are over. I no longer have the time to look into the hairy details of Redis serialisation. PR is welcome.

Thanks.

Nyholm:
> If the Redis adapter doesn't handle serialising/unserialising data types correctly, isn't that a problem for the Redis adapter? I don't really see why my PRS-6 thingy should be compensating for the underlying adapter getting it wrong.

Yes, that is possible. It depends how you want to architecture your Pool, Clients and adapters. If you decide to push that problem to the adapters I would recommend doing integration tests with all adapters instead of a mock, just to make sure the adapters work as well.

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,139 @@
...
+     * Constructor.
+     *
+     * @param string $key
+     * @param mixed $value
+     * @param bool $isHit
+     */
+    public function __construct($key, $value, $isHit)
+    {
+        $this->key = $key;
+        $this->value = $isHit ? $value : null;
```
aequasi: null is a valid value for PSR-6

marc-mabe:
@kynx the redis adapter doesn't handle it wrong. It does what it suppose to do - sending data to redis in a format that redis supports and that's a string only. If you want more you have to add the serializer plugin. That's because not every one need this support and do serializing even if not required would be an useless overhead. That's how and why zend-cache is designed. So if the underlying adapter doesn't support all requirements from psr-6 this wrapper need to implement it on top.

src/Psr/CacheItem.php
kynx: Yes, which is why you need to check isHit() - afaik the value should be null if its not a hit.

kynx:
@marc-mebe yes I had another l

kynx:
Darn phone! Yes, I had another look at this and am completely wrong :(

I'm currently putting together integration tests for each adapter. Once that's done I can start working around the differences.

Nyholm:
About skipping tests: https://github.com/php-cache/integration-tests/pull/16

kynx:
I've added integration tests for each of the storage adapters and am now using the `Serializer` plugin where there are unsupported data types. Can't believe I missed that in the docs to start with :|

I struggled getting phpunit working on a windows VM for too many hours before giving up, so haven't been able to run `test/Psr/WinCacheIntegrationTest.php`. If anyone has a working setup I'd appreciate a check.

Similarly I haven't got the ZendServer adapter tests running yet. I've added a `'minTtl' => 1` to `AbstractZendServer` adapter options there. Hope that's OK - it's in the documentation and just looks like an oversight. 

I've also added some documentation to docs/book.

Looks like `StorageCacheAbstractServiceFactoryTest` is gonna fail due to a change in the ServiceManager dep. Cross my fingers everything else is OK.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,339 @@
...
+            if (method_exists($storage, 'addPlugin')) {
+                $serializer = StorageFactory::pluginFactory('serializer');
+                $storage->addPlugin($serializer);
+            } else {
+                throw new CacheException(sprintf(
+                    "Storage %s requires serialization but does not support 'addPlugin()'",
+                    get_class($storage)
+                ));
+            }
+        }
```
marc-mabe: @kynx I don't thing it's a good idea to actually change the underlying adapter from within a wrapper class. I would be better to hold this information in an object property on initializing the object and directly calling un-/serialize in the wrapper.

src/Storage/Adapter/AbstractZendServer.php
```diff
@@ -216,6 +216,7 @@ protected function internalGetCapabilities()
                         'resource' => false,
                     ],
                     'supportedMetadata'  => [],
+                    'minTtl'             => 1,
```
marc-mabe: please make a separate PR against `master` as this is a bug fix.

marc-mabe:
@kynx @Nyholm What happens if an deferred item expires before pushing it to storage? Is it required to check TTL there, too?

kynx:
It's a good question, and probably one for the FIG ML. And come to that, what if it expires before you call save().

My feeling is that expiry dates from when the item is stored, not when the value is set. But then I'm just a lazy programmer :smile:  

@Nyholm has another interesting one over at https://github.com/php-cache/integration-tests/issues/25. I might be wrong, but implementing anything other than his option C would be non-trivial over here.

But I think it's important to remember these are edge cases. In the 4 years it to the PSR to pass no one thought of them, so I'm not *too* worried.

src/Storage/Adapter/AbstractZendServer.php
kynx: OK, will do.

src/Psr/CacheItemPoolAdapter.php
kynx: Oh, OK. I can see your point.

kynx:
Regarding expiry and deferred items, I guess the relevant part of the specs are:

> TTL - The Time To Live (TTL) of an item is the amount of time between when that item is stored and it is considered stale. The TTL is normally defined by an integer representing time in seconds, or a DateInterval object.

This makes it clear that the TTL is counted from when the item is stored. But...

> Expiration - The actual time when an item is set to go stale. This it typically calculated by adding the TTL to the time when an object is stored, but may also be explicitly set with DateTime object.

I'm not convinced the intention was to treat `expiresAt()` and `expiresAfter()` differently, but will ask the author(s) to explain.

kynx:
I've had a re-think about `expiresAt()` and refactored it so expired items aren't pushed to storage, and aren't a hit. I've also changed the serialization so it's now done by the pool, not be the Serializer plugin.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,360 @@
...
+            $itemTtl = $item->getTtl();
+            if ($itemTtl > 0) {
+                $options->setTtl($itemTtl);
+            }
+
+            // don't save items that have already expired
+            if (is_null($itemTtl) || $itemTtl > 0) {
+                $saved = $this->storage->setItem($item->getKey(), $value);
+                // saved items are a hit? see integration test CachePoolTest::testIsHit()
+                $item->setIsHit(true);
```
marc-mabe: You should check `$saved` before blindly describing the item has hit.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,360 @@
...
+            }
+
+            // reset TTL on adapter, if required
+            $itemTtl = $item->getTtl();
+            if ($itemTtl > 0) {
+                $options->setTtl($itemTtl);
+            }
+
+            // don't save items that have already expired
+            if (is_null($itemTtl) || $itemTtl > 0) {
```
marc-mabe: This should result in exactly the same behavior : `if ($itemTtl >= 0) {`

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,360 @@
...
+            $value = $item->get();
+            if ($this->serializeValues) {
+                $value = serialize($value);
+            }
+
+            // reset TTL on adapter, if required
+            $itemTtl = $item->getTtl();
+            if ($itemTtl > 0) {
+                $options->setTtl($itemTtl);
+            }
```
marc-mabe: Please correct me if I'm wrong but doesn't `0` mean the item never expires and that `null` mean no TTL defined = use the default TTL of the adapter?

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,360 @@
...
+     */
+    public function saveDeferred(CacheItemInterface $item)
+    {
+        if (!$item instanceof CacheItem) {
+            throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
+        }
+
+        // deferred items should always be a 'hit'
+        $item->setIsHit(true);
+        $this->deferred[$item->getKey()] = $item;
```
marc-mabe: Even if the item has been expired already ?

composer.json
```diff
@@ -16,10 +16,13 @@
         "php": ">=5.5",
         "zendframework/zend-stdlib": "~2.5",
         "zendframework/zend-servicemanager": "dev-develop as 2.7.0",
-        "zendframework/zend-eventmanager": "dev-develop as 2.7.0"
+        "zendframework/zend-eventmanager": "dev-develop as 2.7.0",
+        "psr/cache": "^1.0",
+        "psr/log": "^1.0"
```
marc-mabe: the `psr/log` dependency should be removed

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,145 @@
...
+use DateTimeZone;
+use Psr\Cache\CacheItemInterface;
+
+final class CacheItem implements CacheItemInterface
+{
+    private $key;
+    private $value;
+    private $isHit = false;
+    private $expiration = null;
+    private $ttl = null;
```
marc-mabe: Please can you add code comments for each of the properties - thanks

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Returns number of seconds until item expires
+     * @return int|null
+     */
+    public function getTtl()
+    {
+        $ttl = $this->ttl;
+        if ($this->expiration instanceof DateTime) {
+            $now = new DateTime('now', $this->tz);
+            $ttl = $this->expiration->getTimestamp() - $now->getTimestamp();
```
marc-mabe: If you sore the expire time as integer you can calculate the TTL here much faster with `$ttl = $this->expiration - time()`

src/Psr/CacheItemPoolAdapter.php
kynx: In PSR-6-land null means use default TTL. I've added the `getTtl()` method for testing whether an item has expired and to use in this case to set the value when storing. For example,

```
$item->expiresAt(new DateTime(null));
echo $item->getTtl();
// outputs 0 - item has expired
```

src/Psr/CacheItemPoolAdapter.php
kynx: Yeah: `isHit()` checks expiry for items that have had `expiresAt()` set. We don't know that for sure until the time when it's checked.

composer.json
kynx: Will do - need to give travis a kick anyway ;)

src/Psr/CacheItem.php
kynx: Will do.

src/Psr/CacheItem.php
kynx: We need to recalculate so we can find out whether a deferred item has been expired in the meantime.

src/Psr/CacheItemPoolAdapter.php
kynx: See next one - if you've used `expiresAt()` and that date has since passed, `getTtl() <= 0`.

src/Psr/CacheItemPoolAdapter.php
kynx: Good point.

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,145 @@
...
+        return $this->value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isHit()
+    {
+        $ttl = $this->getTtl();
+        return $this->isHit && (is_null($ttl) || $ttl > 0);
```
marc-mabe: Minor performance note: `return $this->isHit && ($ttl = $this->getTtl() === null || $ttl > 0)`

src/Psr/CacheItemPoolAdapter.php
marc-mabe: ah ok I got it

test/Psr/ApcIntegrationTest.php
```diff
@@ -0,0 +1,72 @@
...
+            $this->markTestSkipped('Enable TESTS_ZEND_CACHE_APC_ENABLED to run this test');
+        }
+
+        // set non-UTC timezone
+        $this->tz = date_default_timezone_get();
+        date_default_timezone_set('America/Vancouver');
+
+        // needed on test expirations
+        $this->iniUseRequestTime = ini_get('apc.use_request_time');
+        ini_set('apc.use_request_time', 0);
```
marc-mabe: Does `apc.use_request_time = 1` break PSR-6 specification? So so an exception should be thrown on instantiating in this case :(

kynx: Oh lordy, yes it does. Should've paid more attention to that when I was cut-and-pasting the tests. I've added a check to the constructor and a test.

kynx:
Looks like CI is going to fail until #64 is in.

samsonasik:
setUp() function in tests should use protected modifier in favor of https://github.com/php-cache/integration-tests/pull/31

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,170 @@
...
+        return $this->value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isHit()
+    {
+        $ttl = $this->getTtl();
+        return $this->isHit && ($ttl === null || $ttl > 0);
```
marc-mabe: You should avoid calling `getTtl` if the item is already marked as `isHit === false`

src/Psr/CacheItem.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        $this->value = $value;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAt($expiration = null)
```
marc-mabe: the spec has no default value here `public function expiresAt($expiration);`

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,394 @@
...
+            $isHit = false;
+            try {
+                if ($this->storage->hasItem($key)) {
+                    $value = $this->storage->getItem($key);
+                    $isHit = true;
+
+                    if ($this->serializeValues) {
+                        $value = unserialize($value);
+                    }
+                };
```
marc-mabe: `getItem` should be used with the second argument by reference instead of calling `hasItem` and `getItem` to avoid concurrency issues and for performance reasons as some adapters doesn't natively support `hasItem` (which will be emulated with `getItem`).

Also errors in `unserialize` needs to be handled (see https://github.com/zendframework/zend-serializer/blob/master/src/Adapter/PhpSerialize.php#L71)

```php
// ...
$value = $this->storage->getItem($key, $isHit);
if ($isHit && $this->serializeValues) {
    if ($value === static::$serializedFalse) {
        $value = false;
    } elseif ($value = unserialize($value) === false) {
       // TODO: what should happen if unserialization failed?
    }
}
// ...
```

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,394 @@
...
+                $cacheItems = $this->storage->getItems($keys);
+            } catch (Exception\InvalidArgumentException $e) {
+                throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+            } catch (Exception\ExceptionInterface $e) {
+                $cacheItems = [];
+            }
+
+            foreach ($cacheItems as $key => $value) {
+                if ($this->serializeValues) {
+                    $value = unserialize($value);
```
marc-mabe: Same here : errors in `unserialize` needs to be handled

src/Psr/CacheItemPoolAdapter.php
kynx: Grrr... Yeah, that makes sense, but prophesy sure don't like args-by-reference with `getItem()`. Still looking at what I can do for the tests.

kynx:
A little bit of `Reflection` was required to test unserialization, which feels like a hack, but there we go - prophesy doesn't let me do arg-by-ref.

If a value can't be `unserialize`d it will simply show up as a cache miss. I can't see a straightforward way of reporting the problem to the user that doesn't break PSR-6. But - unless something else is writing to the same storage adapter - I can't see it being more than a sporadic error. Let me know if you've got other ideas.

@samsonasik - thanks for the heads up. Saw your https://github.com/php-cache/integration-tests/pull/31 - fixed here now too.

kynx:
If anyone's interested in Symfony's take on this, check out https://github.com/symfony/symfony/pull/17408

composer.json
```diff
@@ -16,10 +16,12 @@
         "php": ">=5.5",
         "zendframework/zend-stdlib": "~2.5",
         "zendframework/zend-servicemanager": "dev-develop as 2.7.0",
-        "zendframework/zend-eventmanager": "dev-develop as 2.7.0"
+        "zendframework/zend-eventmanager": "dev-develop as 2.7.0",
```
samsonasik: zendframework/zend-eventmanager ^2.6 || ^3.0 ? /cc @weierophinney

composer.json
```diff
@@ -16,10 +16,12 @@
...
+        "psr/cache": "^1.0"
     },
     "require-dev": {
         "zendframework/zend-serializer": "dev-develop as 2.6.0",
+        "cache/integration-tests": "dev-master",
```
aequasi: This may want to change to @stable or `^0.7`

composer.json
marc-mabe: @samsonasik That's not part of this PR - see #64

kynx:
I've rebased against develop and the tests are passing :)

doc/book/psr6.md
```diff
@@ -0,0 +1,132 @@
...
+values like an empty string, `null`, or `false` can be stored. Always check `isHit()` to determine if the item was
+found.
+
+
+## Supported adapters
+
+The PSR-6 specification requires that the underlying storage support time-to-live (TTL), which is set when the
+item is saved. For this reason the following adapters cannot be used: `Dba`, `Filesystem`, `Memory` and `Session`. The 
+`XCache` adapter calculates TTLs based on the request time, not the time the item is actually persisted, which means 
+that it also cannot be used.
```
marc-mabe: I think this is confusing because:
- The filesystem and memory adapter supports TTL but it's based on a stored last-modification-time + runtime TTL setting
  - I'll open an issue for the Memory adapter to change this because this should be a simple change and could also improve performance but it's a BC break.
  - For the filesystem adapter this would be worst because it would be required to read the file before to know it's expired.
- The XCache adapter is not the only adapter calculating TTLs based on request time. APC/APCu also have this behavior if `apc.use_request_time` which is enabled be default.

doc/book/psr6.md
```diff
@@ -0,0 +1,132 @@
...
+            'throw_exceptions' => true,
+        ],
+    ],
+]);
+
+$pool = new CacheItemPoolAdapter($storage);
+```
+
+Note that `throw_exceptions` should always be `true` (the default) or you will not get the correct return values from
+calls on the pool such as `save()`.
```
marc-mabe: Why?

doc/book/psr6.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Supported data types
+
+As per [the specification](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md#data), the 
+following data types can be stored in cache: `string`, `integer`, `float`, `boolean`, `null`, `array`, `object` and be 
+returned as a value with exactly the same type.
+
+Not all adapters can natively store all these types. For instance, Redis stores booleans and integers as a string. Where 
+this is the case *all* values will be automatically run through `serialize()` on save and `unserialize()` on get: you 
+do not need to use a `Zend\Cache\Storage\Plugin\Serializer` plugin.
```
marc-mabe: ... but you can if you like to change the serializer to igbinary.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,419 @@
...
+    public function commit()
+    {
+        $notSaved = [];
+
+        foreach ($this->deferred as &$item) {
+            if (! $this->save($item)) {
+                $notSaved[] = $item;
+            }
+        }
+        $this->deferred = $notSaved;
```
marc-mabe: TODO for later: This could be improved a lot by using the `setItems` method instead but it would increase complexity.

src/Psr/CacheItemPoolAdapter.php
```diff
@@ -0,0 +1,419 @@
...
+        if ($value == static::$serializedFalse) {
+            return ['value' => false, 'isHit' => true];
+        }
+
+        if (false === ($value = unserialize($value))) {
+            return ['value' => null, 'isHit' => false];
+        }
+
+        return ['value' => $value, 'isHit' => true];
+    }
```
marc-mabe: This seems ugly to me. Can we change that to:
```php
private function unserialize($value, & $success)
{
    if ($value === static::$serializedFalse) {
        $success = true;
        return false;
    }

    $value = unserialize($value);
    if ($value === false) {
        $success = false;
        return null;
    }

    $success = true;
    return $value;
}
```
and call like this:
```php
$value = $this->unserialize($value, $isHit);
```

doc/book/psr6.md
kynx: Because we catch the exceptions and return false from these methods in case of errors, as per the spec. If they're not thrown we'll always return true. I'll clarify the docs.

src/Psr/CacheItemPoolAdapter.php
kynx: Yeah, I've thought about this a bit. For optimal performance I'd have to group the items by TTL and send each as a batch after setting the options on the adapter. Maybe one for later.

On a related note we should probably have some kind of (configurable) limit on the number of items that can be deferred before commit() is automatically called to avoid OOM situations in cache-warmers.

src/Psr/CacheItemPoolAdapter.php
kynx: Can do. Some people (like prophesy) hate pass-by-reference. I'm agnostic :)

src/Psr/CacheItemPoolAdapter.php
marc-mabe: Yes this can be done later!

src/Psr/CacheItemPoolAdapter.php
marc-mabe: Yes, pass-by-reference is not the cleanest approach but nowadays IDEs automatically detect that. Returning an associative array and using `extract()` on the other side is much more wired and unreadable.

marc-mabe:
@kynx #80 has been merged into develop now. Please can you update your PR and not allow adapters capability lockOnExpire > 0 thanks!

samsonasik:
@kynx it needs rebase.

kynx:
@samsonasik Yes, apologies, I've been pretty busy recently. I'll look at doing that this weekend.

marc-mabe:
@kynx I did the following changes and merged it now:

* Disallow storages with capability "lock-on-expire" as this violates PSR-6
* calculate TTL from `DateInterval` with just one system call for the current time
* skip PSR integration tests on `ExtensionNotLoadedException`
* added PSR integration test for APCu adapter

Thanks for your work !

kynx:
@marc-mabe Excellent! Thanks for the cleanup, and sorry I faded away. Glad to see this merged!

PR #55, opened 2016-01-03
Fix test canLookupCacheByName

kynx:
This fixes StorageCacheAbstractServiceFactoryTest::canLookupCacheByName() failing due to changes in the ServiceManager::has() functionality when looking up abstract factories (it no longer has a second parameter).


Ocramius:
:+1:

PR #57, opened 2016-01-09
Memcached HHVM compatibility and reduced duplicated code

marc-mabe:


src/Storage/Adapter/Memcached.php
```diff
@@ -515,26 +462,26 @@ protected function internalRemoveItem(& $normalizedKey)
...
         foreach ($normalizedKeys as & $normalizedKey) {
             $normalizedKey = $this->namespacePrefix . $normalizedKey;
         }
 
-        $rsCodes = $memc->deleteMulti($normalizedKeys);
-
         $missingKeys = [];
-        foreach ($rsCodes as $key => $rsCode) {
+        $rsCodes     = $memc->deleteMulti($normalizedKeys);
+        foreach ($rsCodes as $normalizedKey => $rsCode) {
```
Ocramius: Can avoid assigning to `$rsCodes` here

src/Storage/Adapter/Memcached.php
```diff
@@ -256,8 +244,14 @@ protected function internalGetItems(array & $normalizedKeys)
             throw $this->getExceptionByResultCode($memc->getResultCode());
         }
 
+        if (!$result) {
```
Ocramius: should check against `null`, not just with `!` (according to comment below)

marc-mabe: If I check against `NULL` it doesn't catch an empty array returned by standard PHP which than ran into an unnecessary loop. I can check with `if (empty($result))` but that's exactly the same as now.

Ocramius: Ah, that's alright, but the comment below makes it look like this block is only designed to catch HHVM's behavior

marc-mabe: That's why I put the comment into the if-block and not above but a can add another comment above :)

Maks3w: I agree with @Ocramius You describe two behaviors 1. HHVM return type; 2. PHP optimization. but the comment only describe the first.

I think the comment could be removed

src/Storage/Adapter/Memcached.php
```diff
@@ -515,26 +462,26 @@ protected function internalRemoveItem(& $normalizedKey)
...
             return parent::internalRemoveItems($normalizedKeys);
         }
 
-        $memc = $this->getMemcachedResource();
-
         foreach ($normalizedKeys as & $normalizedKey) {
             $normalizedKey = $this->namespacePrefix . $normalizedKey;
         }
 
-        $rsCodes = $memc->deleteMulti($normalizedKeys);
```
Maks3w: What HHVM issue fix?

src/Storage/Adapter/Memcached.php
marc-mabe: updated

src/Storage/Adapter/Memcached.php
marc-mabe: updated

src/Storage/Adapter/Memcached.php
marc-mabe: HHVM doesn't support `deleteMulti`. First I did this difference at this place but decided to do the same as with memcached < 2.0.0 later.

Are this lines better for you now?

Maks3w: I understand the changes above thid line but seems next changes are a microoptimization

marc-mabe: @Maks3w  Sorry but I don't know what you mean. There are two very simple changes only:

- `foreach` directly the result of `deleteMulti` without assigning the result to a variable first. This is what @Ocramius mentioned before with:

> Can avoid assigning to $rsCodes here

- renaming the variable `$key` to `$normalizedKey` as it holds the same value as given to `$memc->deleteMulti($normalizedKeys)`

marc-mabe:
@Maks3w @Ocramius Is this PR ready for merge now?

marc-mabe:
@Maks3w @Ocramius ping

Issue #58, opened 2016-01-09
Fatal error storing objects in xcache adapter


kynx:
> Fatal error: xcache_set(): Objects cannot be stored in the variable cache. Use serialize before xcache_set in /Users/matt/www/httpd/zend-cache/src/Storage/Adapter/XCache.php on line 348

As the [docs](https://xcache.lighttpd.net/wiki/XcacheApi) say:

> **Warning** : At the moment, It is not possible to store resources, callbacks or objects using xcache_* functions.

Looks the adapter needs to have "object" => false in it's supportedDatatypes options and also throw an exception instead of fatal error. I'm using xcache 3.2.0.

BTW, xcache > 3.1.0 does support CLI, so there's no need to skip the tests for stuff like this. I'll do a PR.

kynx:
Oh lordy, looks like namespaces are broken too:

```
$ TESTS_ZEND_CACHE_XCACHE_ENABLED=1 phpunit test/Storage/Adapter/XCacheTest.php 
PHPUnit 4.8.9 by Sebastian Bergmann and contributors.

.................S.......F.FF...............................S.... 65 / 72 ( 90%)
..SSS..

Time: 19.46 seconds, Memory: 16.50Mb

There were 3 failures:

1) ZendTest\Cache\Storage\Adapter\XCacheTest::testGetMetadatasWithEmptyNamespace
Failed asserting that 0 is identical to 2.

/Users/matt/www/httpd/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:391
/Users/matt/www/httpd/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:407

2) ZendTest\Cache\Storage\Adapter\XCacheTest::testSetGetHasAndRemoveItemWithoutNamespace
Failed asserting that null matches expected 'value'.

/Users/matt/www/httpd/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:423

3) ZendTest\Cache\Storage\Adapter\XCacheTest::testSetGetHasAndRemoveItemsWithoutNamespace
Failed asserting that an array has the key 'key1'.

/Users/matt/www/httpd/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:446

```

marc-mabe:
Ohhh it looks like they removed the serializing support and they have problems keeping the extension up-to-date. (Could be because of less of interest as with opcache and apcu exists a much more common approach)

[...]

PR #59, opened 2016-01-09
Fix for #58

kynx:
* XCache adapter now checks version > 3.1.0 when running under CLI
* fixed broken internalSetItem() with empty namespace

Don't got the chops to get travis building the xcache extension so the tests can run - doesn't look like it's avail from pecl (I get mine from brew). Anyone got any ideas?

### DIFF
--- a/src/Storage/Adapter/XCache.php
+++ b/src/Storage/Adapter/XCache.php
@@ -54,9 +54,9 @@ class XCache extends AbstractAdapter implements
             throw new Exception\ExtensionNotLoadedException('Missing ext/xcache');
         }
 
-        if (PHP_SAPI == 'cli') {
+        if (PHP_SAPI == 'cli' && version_compare(phpversion('xcache'), '3.1.0') < 0) {
             throw new Exception\ExtensionNotLoadedException(
-                "ext/xcache isn't available on SAPI 'cli'"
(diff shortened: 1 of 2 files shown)


kynx:
Ach, since github doesn't link from the titles, this addresses #58

src/Storage/Adapter/XCache.php
```diff
@@ -341,10 +341,17 @@ protected function internalSetItem(& $normalizedKey, & $value)
     {
         $options     = $this->getOptions();
         $namespace   = $options->getNamespace();
-        $prefix      = ($options === '') ? '' : $namespace . $options->getNamespaceSeparator();
+        $prefix      = ($namespace === '') ? '' : $namespace . $options->getNamespaceSeparator();
```
marc-mabe: good catch :+1:

src/Storage/Adapter/XCache.php
```diff
@@ -341,10 +341,17 @@ protected function internalSetItem(& $normalizedKey, & $value)
...
         $internalKey = $prefix . $normalizedKey;
         $ttl         = $options->getTtl();
 
+        if (is_object($value) || is_resource($value) || is_callable($value)) {
```
marc-mabe: What does `is_callable` catch here for you as `is_callable` could be simple string, too.

kynx:
Fixed in #65

PR #60, opened 2016-01-10
Rename StorageCacheAbstractserviceFactory::canCreateServiceWithName

kynx:
…to canCreate() so it works with latest ServiceManager


kynx:
Looks like this will be addressed in #64

PR #62, opened 2016-01-12
Fix StorageCacheAbstractServiceFactory to support the updated AbstractFactoryInterface

MidnightDesign:


weierophinney:
Accomplished with #64 / #68 / #69, which also update the component for full forwards-compatibility with zend-stdlib, zend-eventmanager, and zend-servicemanager v3.

PR #63, opened 2016-01-14
Implemented the ClearByNamespace interface in the Redis adapter

aimfeld:
When switching from APC to Redis, I noticed that the Redis adapter did not implement the ClearByNamespace interface. The provided implementation works well for us and is based on the code  in `Apc::clearByNamespace()` and `Redis::clearByPrefix()`.


marc-mabe:
@aimfeld Your PR looks good but I'm waiting for #64 before. Thanks!

PR #65, opened 2016-01-14
enable xcache tests on travis (based on #59)

marc-mabe:
this PR is based on #59 and addresses #58 

@kynx It took my some hours to make it work in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,10 @@ env:
     - TESTS_ZEND_CACHE_REDIS_PORT=6379
     - TESTS_ZEND_CACHE_REDIS_PASSWORD=''
     - TESTS_ZEND_CACHE_REDIS_DATABASE=0
+    - TESTS_ZEND_CACHE_XCACHE_ENABLED=true
+    - TESTS_ZEND_CACHE_XCACHE_ADMIN_AUTH=true
+    - TESTS_ZEND_CACHE_XCACHE_ADMIN_USER=travis
+    - TESTS_ZEND_CACHE_XCACHE_ADMIN_PASS=test
 
 matrix:
   fast_finish: true
@@ -39,10 +43,12 @@ matrix:
       env:
         - EXECUTE_CS_CHECK=true
         - PECL_INSTALL_APCU='apcu-4.0.8'
+        - INSTALL_XCACHE=true
     - php: 5.6
       env:
         - EXECUTE_TEST_COVERALLS=true
         - PECL_INSTALL_APCU='apcu-4.0.8'
+        - INSTALL_XCACHE=true
     - php: 7
       env:
         - PECL_INSTALL_APCU='apcu'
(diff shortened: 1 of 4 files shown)


kynx:
@marc-mabe Looks good! I'll have to get my head round travis properly sometime.

Maks3w:
Why PHP 7 is excluded?

marc-mabe:
@Maks3w because the latest xcache version is not PHP-7 compatible

src/Storage/Adapter/XCache.php
```diff
@@ -434,7 +441,7 @@ protected function internalGetCapabilities()
                         'double'   => true,
                         'string'   => true,
                         'array'    => true,
-                        'object'   => 'object',
+                        'object'   => false,
```
Maks3w: Can you document why this change is needed? I wonder if some kind of bc break introduced by Xcache and so a minimum version is required

marc-mabe: mh you are right but the bc break is in xcache because as far as I know it worked with older versions and results in a fatal error now. What we could do here is to find out at which version this feature has been removed and set the supported datatypes according to it.
-> Than in my opinion we no longer have a bc break here.

Maks3w: We "may" break the minimum version supported

I imagine this change may deail with one of these possible scenarios.

a. Minimum Xcached version is raised and older versions are not longer supported
b. Minimum Xcached version is raised and CLI assertion is not longer supported
c. Set the setting based in the installed Xcached version for keep supporting previous versions and new ones
d. Old Xcached versions are compatible with these change and this change is OK
e. XCache adapter was not working with any version from far ago so there is no any kind of bc break

Maks3w:
Then we should "TESTS_ZEND_CACHE_XCACHE_ENABLED=false" for PHP 7 instead add a new env var (INSTALL xcache). We already have a var for this!

src/Storage/Adapter/XCache.php
marc-mabe: @Maks3w  I grepped the history of xcache but even in the first available commit on github from 2012 this feature was not supported by xcache - see https://github.com/lighttpd/xcache/blob/6663fcfd7a174eb972618c4c70c55b04b5b75024/mod_cacher/xc_cacher.c#L2796
-> So for me this is a bugfix

marc-mabe:
@Maks3w I prefer the set all `TESTS_ZEND_CACHE_*_ENABLED` variables to true in travis as this enables testing that the right exception will be thrown an no fatal error happens if an extension is missing.

The variables `TESTS_ZEND_CACHE_*_ENABLED` only exists because it writes test data into a storage and we don't know if we break someones applications or overwrite something as we don't know in which environments the tests are running.

In an optimal word we should enables all adapters on travis and let the tests run with and without the required extensions installed but this would slow down travis a lot.

.travis.yml
```diff
@@ -63,6 +69,7 @@ before_install:
 
 install:
   - travis_retry composer install --no-interaction --ignore-platform-reqs
+  - if [[ $INSTALL_XCACHE == 'true' ]]; then git clone https://github.com/lighttpd/xcache.git && cd xcache && git checkout $(git rev-list --tags --max-count=1) && phpize && ./configure --enable-xcache && make install && cd .. && phpenv config-add .ci/xcache.ini || cd .. ; fi
```
Maks3w: I've hear now is possible use multi-line statements https://github.com/omnia-md/conda-dev-recipes/blob/master/.travis.yml

marc-mabe: thanks for the tip - updated all as multi-line

marc-mabe:
@Maks3w It's using `pwd=$(pwd)` now as with `pushd` and `popd` it's nearly the same complicated as before. (It needs to make sure if one of the commands failed in the chain that it can go back the the original directory) I hope it's simpler to understand / read now

Maks3w:
@marc-mabe It's ok for me. However I would prefer for future changes the default policy in the test matrix should be enable the tests (added as global env var) and disable for specific versions (like PHP7)

Issue #72, opened 2016-02-12
array_key_exists() error when updating from 2.5.3 to 2.6


avasa:
Error: 

array_key_exists(): The first argument should be either a string or an integer

Calling:

```php
$config['cache'] = 'adapter' => [
    'name' => 'memcached',
            'options' => [
                'servers' => [
                    [
                        'host' => 'localhost',
                        'port' => '11211',
                    ],
                ],
            ],
        ];

$cache = StorageFactory::factory($config['cache']);
```

Stack:

```
Whoops\Exception\ErrorException: array_key_exists(): The first argument should be either a string or an integer
  File "/Users/user/dev/myapp/public/index.php", line 10, in null
    $app->run();
  File "/Users/user/dev/myapp/vendor/zendframework/zend-mvc/src/Application.php", line 314, in run
    $result = $events->trigger(MvcEvent::EVENT_DISPATCH, $event, $shortCircuit);
  File "/Users/user/dev/myapp/vendor/zendframework/zend-eventmanager/src/EventManager.php", line 214, in trigger
    return $this->triggerListeners($event, $e, $callback);
  File "/Users/user/dev/myapp/vendor/zendframework/zend-eventmanager/src/EventManager.php", line 490, in triggerListeners
    $responses->push(call_user_func($listenerCallback, $e));
  File "/Users/user/dev/myapp/vendor/zendframework/zend-mvc/src/DispatchListener.php", line 93, in onDispatch
    $return = $controller->dispatch($request, $response);
  File "/Users/user/dev/myapp/vendor/zendframework/zend-mvc/src/Controller/AbstractController.php", line 118, in dispatch
    });
  File "/Users/user/dev/myapp/vendor/zendframework/zend-eventmanager/src/EventManager.php", line 214, in trigger
    return $this->triggerListeners($event, $e, $callback);
  File "/Users/user/dev/myapp/vendor/zendframework/zend-eventmanager/src/EventManager.php", line 490, in triggerListeners
    $responses->push(call_user_func($listenerCallback, $e));
  File "/Users/user/dev/myapp/module/myapp/src/Controller/CliController.php", line 44, in onDispatch
    $viewModel = parent::onDispatch($e);
  File "/Users/user/dev/myapp/vendor/zendframework/zend-mvc/src/Controller/AbstractActionController.php", line 82, in onDispatch
    $actionResponse = $this->$method();
[...]

Issue #71, opened 2016-02-12
Broken Events v2 compatibility


bacinsky:
I think this just can`t work:
https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/AbstractAdapter.php#L193
https://github.com/zendframework/zend-eventmanager/blob/release-2.6.2/src/EventManager.php#L57

BTW, I'm not offending, just complaining, with the v3 oncoming this backward compatibility issues got worse directly, into masters. You know I can`t be happy with things like that. Spending time finding BC breaks in the ZF2 instead of doing work for my clients. I really appreciate your work but this is sad for me that the ZF2 is going like this.

pobidowski:
Confirm

bgaillard:
Hi we encountered the same problem after upgrading Zend-cache from version `2.5.3` to version `2.6.0`.

simonbaudry:
Confirm as well on the version 2.6.0

weierophinney:
@bacinsky We are trying very hard to test against both v2 and v3 APIs to ensure things going break; this particular line clearly didn't have a test case covering it. I'll get a fix our shortly, as there is a way to make this work on both versions. 

Thanks for the report!

neoglez:
https://github.com/zendframework/zend-cache/issues/72 is a direct consequence of this break.

weierophinney:
Those affected by this issue: can you please test the patch from #73? Tests pass against both event manager versions, and I added tests explicitly for the lazy-loading aspect, which should resolve the issue. However, I'd like verification from somebody affected.

bgaillard:
Hi @weierophinney and thanks for your reactivity on this, I tested the patch and it seems to work.

quintanilhar:
Hi @weierophinney, I also tested this patch and worked fine.

Waiting for the merge :)

Thank you!

PR #73, opened 2016-02-12
Use v2-compatible EventManager instantiation when lazy-loading

weierophinney:
Per #71, `getEventManager()`, when lazy-loading an `EventManager`, was using the v3 syntax, instead of v2-compatible syntax, causing identifiers to break under v2 (as they are now injected with a `SharedEventManager` instance instead).

This patch:

- Adds a compatiblity test to ensure that the usage works on both v2 and v3.
- Fixes the lazy-loading code to instantiate and then inject identifiers.

This also addresses #72, as the problem presented in that report is due to the fact that the `SharedEventManager` was injected as an identifier, and identifiers are only allowed to be strings or integers.

### DIFF
--- a/test/Storage/Adapter/EventManagerCompatibilityTest.php
+++ b/test/Storage/Adapter/EventManagerCompatibilityTest.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ * @package   Zend_Cache
+ */
+
+namespace ZendTest\Cache\Storage\Adapter;
+
+use PHPUnit_Framework_TestCase as TestCase;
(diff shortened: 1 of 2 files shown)


test/Storage/Adapter/EventManagerCompatibilityTest.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $events = $this->adapter->getEventManager();
+        $this->assertInstanceOf(EventManager::class, $events);
+        return $events;
+    }
+
+    /**
+     * @depends testCanLazyLoadEventManager
+     */
+    public function testLazyLoadedEventManagerIsInjectedProperlyWithDefaultIdentifiers(EventManager $events)
```
Ocramius: To be sure, I'd also add an `@coversNothing` test that triggers an event (any event) on the event manager and on its attached shared event manager (integration test)

weierophinney: Ugh, missed this comment!

That said, the other tests are already testing triggering of events, and that works between both versions. The difference, however, is that those inject an EM instance (or mock) to do so, so they're primarily testing that the correct calls are made. I *did*, however, have issues crop up in testing yesterday when completing the migration patch due to compatibility issues, which I was able to fix. I'll see if I can come up with an additional test to be on the safe-side, though.

PR #82, opened 2016-02-19
Bugfix/clear by tag on filesystem storage while files are already deleted

### LINKED ISSUE #44: clearByTags in Filesystem adapter crashes when .tag file deleted during GlobIterator loop
Hi,

we see the following error when using Filesystem storage adapter and deleting a large set of object from cache with `clearByTags()` method.

```
// when file_locking is off in File Zend/Cache/Storage/Adapter/Filesystem.php in line 1363
file_get_contents(my_cache_dir/my_cached_object_4711.tag): failed to open stream: No such file or directory
// when file_locking is on in File Zend/Cache/Storage/Adapter/Filesystem.php in line 1327
fopen(my_cache_dir/my_cached_object_4711.tag): failed to open stream: No such file or directory
```

It is not that easy to reproduce but occurs as follows:

- the `GlobIterator` in `clearByTags()` looks for all .tag files which holds the cached item tags information
- while looping through these items, another process deletes one item (and its .tag file) which is also prior found by the `GlobIterator`
- `clearByTags` triggers an error at `$this->getFileContent($pathname)` because `fopen()` or `file_get_contents` try to open a non existing file (not existing anymore)

Should be simple be avoided by adding
```php
if (!file_exists($file)) { // or !$glob->valid()?
    continue;
}
```

to the foreach loop in `clearByTags()`, but is this right?

I also can submit a PR, but until now i do not have found a way to reproduce for a unit test

cheers

skors:
This PR fixes the behavior on the special case that file content is read while tag file is deleted

More description in #44 

@marc-mabe is this kind of try...catch solution you had in mind?

Merging this PR closes #44

### DIFF
--- a/src/Storage/Adapter/Filesystem.php
+++ b/src/Storage/Adapter/Filesystem.php
@@ -331,15 +331,11 @@ class Filesystem extends AbstractAdapter implements
             try {
                 $diff = array_diff($tags, explode("\n", $this->getFileContent($pathname)));
             } catch (Exception\RuntimeException $exception) {
-                if ($exception->getPrevious()
-                    && strpos(
-                        $exception->getPrevious()->getMessage(),
-                        'failed to open stream: No such file or directory'
-                    ) !== false
-                ) {
+                // ignore missing files because of possible raise conditions
+                // e.g. another process already deleted that item
+                if (!file_exists($pathname)) {
                     continue;
                 }
-
                 throw $exception;
             }
 
--- a/src/Storage/Adapter/Filesystem.php
+++ b/src/Storage/Adapter/Filesystem.php
@@ -328,7 +328,20 @@ class Filesystem extends AbstractAdapter implements
         $glob = new GlobIterator($path, $flags);
 
         foreach ($glob as $pathname) {
-            $diff = array_diff($tags, explode("\n", $this->getFileContent($pathname)));
+            try {
+                $diff = array_diff($tags, explode("\n", $this->getFileContent($pathname)));
+            } catch (Exception\RuntimeException $exception) {
+                if ($exception->getPrevious()
+                    && strpos(
+                        $exception->getPrevious()->getMessage(),
+                        'failed to open stream: No such file or directory'
+                    ) !== false
+                ) {
+                    continue;
+                }
+
+                throw $exception;
+            }
 
             $rem  = false;
             if ($disjunction && count($diff) < $tagCount) {
--- a/test/Storage/Adapter/FilesystemTest.php
+++ b/test/Storage/Adapter/FilesystemTest.php
@@ -334,6 +334,10 @@ class FilesystemTest extends CommonAdapterTest
      */
     public function testClearByTagsWithoutLocking()
     {
+        if (!function_exists('pcntl_fork') || !function_exists('posix_kill')) {
+            $this->markTestSkipped('Missing pcntl_fork and/or posix_kill');
+        }
+
         // create cache items
         $this->_storage->getOptions()->setDirLevel(0);
         $this->_storage->getOptions()->setFileLocking(false);
@@ -378,6 +382,10 @@ class FilesystemTest extends CommonAdapterTest
      */
     public function testClearByTagsWithLocking()
     {
+        if (!function_exists('pcntl_fork') || !function_exists('posix_kill')) {
+            $this->markTestSkipped('Missing pcntl_fork and/or posix_kill');
+        }
+
         // create cache items
         $this->_storage->getOptions()->setDirLevel(0);
         $this->_storage->getOptions()->setFileLocking(true);
--- a/test/Storage/Adapter/FilesystemTest.php
+++ b/test/Storage/Adapter/FilesystemTest.php
@@ -347,15 +347,9 @@ class FilesystemTest extends CommonAdapterTest
         $this->_storage->setTags('b_key', ['a_tag', 'b_tag']);
         $this->_storage->setItem('c_key', 'c_value');
         $this->_storage->setTags('c_key', ['a_tag', 'c_tag']);
-        //$this->_storage->getTags('a_key');
 
-        //unlink($this->_tmpCacheDir.'/zfcache-a_key.tag');
-
-        //var_dump($this->_storage->getTags('a_key'));
-
-        // tempFile used for communication between parent and child.
         $tempFile = tmpfile();
-        // Split into two..
+        
         $pid = pcntl_fork();
         if ($pid == -1) {
             $this->fail('pcntl_fork() failed');
--- a/test/Storage/Adapter/FilesystemTest.php
+++ b/test/Storage/Adapter/FilesystemTest.php
@@ -411,4 +411,3 @@ class FilesystemTest extends CommonAdapterTest
         }
     }
 }
-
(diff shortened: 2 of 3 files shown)


src/Storage/Adapter/Filesystem.php
```diff
@@ -328,7 +328,20 @@ public function clearByTags(array $tags, $disjunction = false)
...
+                    && strpos(
+                        $exception->getPrevious()->getMessage(),
+                        'failed to open stream: No such file or directory'
+                    ) !== false
+                ) {
+                    continue;
+                }
+
+                throw $exception;
+            }
```
DCMNMarc: This looks very error prone. I think it would be better to test if the file exists instead of parsing an exception message.

```php
try {
    // ...
} catch (Exception\RuntimeException $e) {
    // ignore missing files because of possible raise conditions
    // e.g. another process already deleted that item
    if (!file_exists($pathname)) {
        continue;
    }
    throw $e;
}
```

test/Storage/Adapter/FilesystemTest.php
```diff
@@ -328,4 +328,86 @@ public function testClearByPrefixWithUnexpectedDirectory()
...
+        //unlink($this->_tmpCacheDir.'/zfcache-a_key.tag');
+
+        //var_dump($this->_storage->getTags('a_key'));
+
+        // tempFile used for communication between parent and child.
+        $tempFile = tmpfile();
+        // Split into two..
+        $pid = pcntl_fork();
+        if ($pid == -1) {
+            $this->assertEquals('shit went down. [failed to fork]', 'no shit went down');
```
DCMNMarc: $this->fail('pcntl_fork() failed');

test/Storage/Adapter/FilesystemTest.php
```diff
@@ -328,4 +328,86 @@ public function testClearByPrefixWithUnexpectedDirectory()
...
+
+            $this->_storage->clearByTags(['a_tag'], true);
+
+            fseek($tempFile, 0);
+            $messageFromChild = fread($tempFile, 1024);
+            $this->assertEquals($messageFromChild, 'deleted');
+            fclose($tempFile);
+        } else {
+            usleep(150000);
+            unlink($this->_tmpCacheDir.'/zfcache-c_key.tag');
```
DCMNMarc: You can get the filename by `$this->_storage->getMetadata('c_key')` but it would be much cleaner if you are using the storage adapter directly to delete the item `$this->_storage->removeItem('c_key')`.

test/Storage/Adapter/FilesystemTest.php
```diff
@@ -328,4 +328,86 @@ public function testClearByPrefixWithUnexpectedDirectory()
...
+        //$this->_storage->getTags('a_key');
+
+        //unlink($this->_tmpCacheDir.'/zfcache-a_key.tag');
+
+        //var_dump($this->_storage->getTags('a_key'));
+
+        // tempFile used for communication between parent and child.
+        $tempFile = tmpfile();
+        // Split into two..
+        $pid = pcntl_fork();
```
DCMNMarc: All special functions needs to be tested before using else this test will run into fatals on some environments:
```php
if (!function_exists('pcntl_fork') || !function_exists('posix_kill')) {
    $this->markTestSkipped('Missing pcntl_fork and/or posix_kill');
}
```

src/Storage/Adapter/Filesystem.php
skors: @DCMNMarc thanks, that was unclear to me.

skors: @DCMNMarc i've changed it. but wouldn't this behave same like you commented in https://github.com/zendframework/zend-cache/issues/44#issuecomment-155558236 ?

test/Storage/Adapter/FilesystemTest.php
skors: You are right, i was too focused on just the tag file...

marc-mabe:
@skors Much thanks for this fix and unittest!

Today I had time play a little with it and was able to reduce the wait timings and removed the temporary file the tests still works (or fails) as expected.

This way the test should also fail if the process wait timings are not proper or even if something changed in the filesystem adapter which breaks the hack for delay `unlink`.

But I still doesn't understand why `posix_kill(posix_getpid(), SIGTERM);` is required.

Now it looks like the following:
```php
    /**
     * @runInSeparateProcess
     */
    public function testClearByTagsWithoutLocking()
    {
        if (!function_exists('pcntl_fork') || !function_exists('posix_kill')) {
            $this->markTestSkipped('Missing pcntl_fork and/or posix_kill');
        }

        // create cache items
        $this->_storage->getOptions()->setDirLevel(0);
        $this->_storage->getOptions()->setFileLocking(false);
        $this->_storage->setItems([
            'a_key' => 'a_value',
            'b_key' => 'b_value',
            'other' => 'other',
        ]);
        $this->_storage->setTags('a_key', ['a_tag']);
        $this->_storage->setTags('b_key', ['a_tag']);

        $pidChild = pcntl_fork();
        if ($pidChild == -1) {
            $this->fail('pcntl_fork() failed');
        } elseif ($pidChild) {
            // The parent process
            // Slow down unlink function and start removing items.
            // Finally test if the item not matching the tag was removed by the child process.
            require __DIR__ . '/TestAsset/FilesystemDelayedUnlink.php';
            $this->_storage->clearByTags(['a_tag'], true);
            $this->assertFalse($this->_storage->hasItem('other'));
        } else {
            // The child process:
            // Wait to make sure the parent process has started determining files to unlink.
            // Than remove one of the items the parent process should remove and another item for testing.
            usleep(10000);
            $this->_storage->removeItems(['b_key', 'other']);
            posix_kill(posix_getpid(), SIGTERM);
        }
    }

    /**
     * @runInSeparateProcess
     */
    public function testClearByTagsWithLocking()
    {
        if (!function_exists('pcntl_fork') || !function_exists('posix_kill')) {
            $this->markTestSkipped('Missing pcntl_fork and/or posix_kill');
        }

        // create cache items
        $this->_storage->getOptions()->setDirLevel(0);
        $this->_storage->getOptions()->setFileLocking(true);
        $this->_storage->setItems([
            'a_key' => 'a_value',
            'b_key' => 'b_value',
            'other' => 'other',
        ]);
        $this->_storage->setTags('a_key', ['a_tag']);
        $this->_storage->setTags('b_key', ['a_tag']);

        $pidChild = pcntl_fork();
        if ($pidChild == -1) {
            $this->fail('pcntl_fork() failed');
        } elseif ($pidChild) {
            // The parent process
            // Slow down unlink function and start removing items.
            // Finally test if the item not matching the tag was removed by the child process.
            require __DIR__ . '/TestAsset/FilesystemDelayedUnlink.php';
            $this->_storage->clearByTags(['a_tag'], true);
            $this->assertFalse($this->_storage->hasItem('other'));
        } else {
            // The child process:
            // Wait to make sure the parent process has started determining files to unlink.
            // Than remove one of the items the parent process should remove and another item for testing.
            usleep(10000);
            $this->_storage->removeItems(['b_key', 'other']);
            posix_kill(posix_getpid(), SIGTERM);
        }
    }
```

marc-mabe:
closed by bb8a75c

Thanks!

PR #84, opened 2016-02-23
Adds Redis sIsMember and sAdd Taggable capabilities [POC]

### LINKED ISSUE #83: Redis sIsMember or sContains?
Maybe there is a new way of doing this that I am unaware of but I don't see why [this would not be a feature within the redis adapter?](https://github.com/phpredis/phpredis#sismember-scontains)

#### Example:
```php
$redis->sAdd('key1' , 'member1');
$redis->sAdd('key1' , 'member2');
$redis->sAdd('key1' , 'member3'); /* 'key1' => {'member1', 'member2', 'member3'}*/

$redis->sIsMember('key1', 'member1'); /* TRUE */
$redis->sIsMember('key1', 'memberX'); /* FALSE */
```

jeremymills:
#### PROOF OF CONCEPT

- Implementation could also just be added to redis alone, as apposed to the Abstract Adapter ?

fixes #83

example request:

```php
$this->redis->sAdd('key', 'value1');
$this->redis->sAdd('key', 'value2');
$this->redis->sAdd('key', 'value3');

$this->sIsMember('key', 'value1'); // returns true
$this->sIsMember('key', 'value2'); // returns true
$this->sIsMember('key', 'value6'); // returns false
```

### DIFF
--- a/src/Storage/Adapter/AbstractAdapter.php
+++ b/src/Storage/Adapter/AbstractAdapter.php
@@ -321,53 +321,6 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
     /* reading */
 
     /**
-     * Returns if value is a member of the set stored at key.
-     *
-     * @param  string  $key
-     * @param  string  $value
-     * @return int 1 if the element is a member of the set, 0 if not or key doesn't exist
-     * @throws Exception\ExceptionInterface
-     *
-     * @triggers sIsMember.pre(PreEvent)
-     * @triggers sIsMember.post(PostEvent)
-     * @triggers sIsMember.exception(ExceptionEvent)
-     */
-    public function sIsMember($key, $value)
-    {
-        if (!$this->getOptions()->getReadable()) {
-            return false;
-        }
-
-        $this->normalizeKey($key);
-        $args = new ArrayObject([
-            'key'   => & $key,
-            'value' => & $value
-        ]);
-
-        try {
-            $eventRs = $this->triggerPre(__FUNCTION__, $args);
-
-            $result = $eventRs->stopped()
-                ? $eventRs->last()
-                : $this->internalSIsMember($args['key'], $args['value']);
-
-            return $this->triggerPost(__FUNCTION__, $args, $result);
-        } catch (\Exception $e) {
-            $result = 0;
-            return $this->triggerException(__FUNCTION__, $args, $result, $e);
-        }
-    }
-
-    /**
-     * Internal method to check if value is a member of the set stored at key.
-     *
-     * @param  string $normalizedKey
-     * @param $value
-     * @return int 1 if the element is a member of the set, 0 if not or key doesn't exist
-     */
-    abstract protected function internalSIsMember(& $normalizedKey, & $value);
-
-    /**
      * Get an item.
      *
      * @param  string  $key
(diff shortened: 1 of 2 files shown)


skors:
@jeremymills am i right that this addition could be used to make Redis Cache Adapter with taggable support?

jeremymills:
@skors yes, as well as, if [POC] is approved,
```php
function sMembers($key) {}
```
will return the entire result set of (in this example) `key`

skors:
Nice! :+1: 

Why didn't you implement TaggableInterface in Zend\Cache\Storage\Adapter\Redis?

jeremymills:
@skors `sMembers` and `s*` methods are all derived from redis itself, what I am doing is simply adding in the "piggyback" to redis core in the adapter, so there is no need to implement the taggable interface.

skors:
I would suggest to implement TaggableInterface in RedisAdapter and would avoid adding these redis specific methods to AbstractAdapter, because all other adapters will not be compatible to these methods. in the redis adapter the *Tag* Methods should implement the logic to be suitable with the redis sRem and sMember* Methods

jeremymills:
Yeah I went ahead and implemented the TaggableInterface in my previous commit above your comment, the one issue I see is with `clearByTags` I'll take a closer look at my implementation

jeremymills:
I may be wrong but I don't believe this test is proper.. if you remove one tag from a set and that set has 2 or more tags, why would it delete the entire set?

https://github.com/zendframework/zend-cache/blob/master/test/Storage/Adapter/CommonAdapterTest.php#L1124-L1125

DCMNMarc:
@jeremymills The test is working fine!

* `setTags($key, $tags)`: should set all given tags for the given key of an existing item. If the given `$tags` is an empty array only the referenced tags for an existing item should be removed, the item itself should not be touched.
* `getTags($key)`:  should return all tags for the given key of an existing item.
* `clearByTags($tags, false)`: should remove all items where all given tags match
* `clearByTags($tags, true)`: should remove all items where at least one of the given tags match.

A new method  `isTag($key, $tag)` doesn't make sense because it only checks if one tag is matching but the `TaggableInterface` works with more than one key. So the only method that could make sense would be `hasTags($key, array $tags, $disjunction = false)` but it would be a BC break because of a changed interface.

jeremymills:
Ok thanks @DCMNMarc, I will update the pr

jeremymills:
@DCMNMarc, I believe the implementation of `Redis::sIsMember` would be extremely useful. I have taken your advice and added a `hasTags($key, array $tags, $disjunction = false)` within the Redis Adapter and not the Interface, for the time being, as it would be a BC break.

DCMNMarc:
@jeremymills
The problem with your current implementation is that you are using the tags Interface to store items as a set. This is wrong. Tags are designed to tag existing items (metadata)

The following should work:
'''
$cache->setItem("test", "value");
$cache->setTags("test", ["tag"]);
$cache->getItem("test"); // value
$cache->getTags("test"); // [tag]
$cache->clearByTags(["tag"]);
$cache->getItem("test"); // null
'''

jeremymills:
@DCMNMarc got it thank you! I've been crazy busy and haven't had a chance to update anything as of late, I will take you're suggestion and update it

marc-mabe:
I'm closing this now as there is no update and I still feel unconvinced with this.

PR #91, opened 2016-03-18
prioritized adapter options

marc-mabe:
fixes zendframework/zf2#6381

PS: This is based on #90 as it requires the `AdapterOptionsTest` file


marc-mabe:
@Ocramius Did you had something like that in mind ?

PR #94, opened 2016-04-11
Allow standalone usage as a component

weierophinney:
This patch accomplishes a number of tasks in order to better faciliate usage of the package as a standalone component, in both Expressive and zend-mvc applications. Specifically:

- Updates the `PatternPluginManager` to accept `$options`, cast them to `PatternOptions` instances, and inject them in created plugins. Doing so allows moving that logic out of the `PatternFactory`, and better facilitates re-use of the plugin manager.
- Updates the `StorageCacheFactory` and `StorageCacheAbstractServiceFactory` to seed the `StorageFactory` with plugin manager/adapter plugin manager instances pulled from the container provided at invocation, if available (falling back to creating new instances with the provided container as the parent container).
- Adds factories for each of the plugin managers (pattern plugins, storage plugins, and storage adapter plugins).

These last two changes allow configuring plugin managers with custom plugins, which then become available to the cache storage when requested.

Finally, this patch adds:

- `Zend\Cache\ConfigProvider`, which maps the abstract service factory and all plugin managers by default.
- `Zend\Cache\Module`, which does the same, but for zend-mvc contexts.

These changes are backwards compatible, but will make the component a better citizen of zend-mvc v3 and Expressive applications.


weierophinney:
@marc-mabe This is part of the ZF3 initiative. We're setting up components to be installed individually, and [zend-component-installer](https://zendframework.github.io/zend-component-installer/) will then register them as modules in zend-mvc applications, or within the config merger in Expressive applications. By doing so, developers will not need to manually register default factory configuration, simplifying documentation and setup.

If you're okay with this, I'll merge and tag a new 2.7.0 release, which will also include the APCu changes you've already merged.

PR #100, opened 2016-05-25
We don't need call info method every time we want redis resource

alireza-m:


marc-mabe:
@alireza-m Thank you for finding this!
But there is one possibility the version information would not get detected.
-> In case you add an existing and already connected redis resource to the resource manager - see https://github.com/alireza-m/zend-cache/blob/ede35a3dc6215decf88f23feaefac7438195e3cf/src/Storage/Adapter/RedisResourceManager.php#L337

The logic of this `initialized` flag is really strange on here and should be fixed.

Moln:
But  the `$resource['version']` can be set only once.
See  https://github.com/zendframework/zend-cache/issues/119

PR #74, opened 2016-02-12
Travis: PHP-7 is no longer allowed to fail

marc-mabe:
But something with the Memcached adapter seems to be wrong:
```
There were 4 failures:

1) ZendTest\Cache\Storage\Adapter\MemcachedTest::testCheckAndSetItem

Failed asserting that null is not null.

/home/travis/build/zendframework/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:802

2) ZendTest\Cache\Storage\Adapter\MemcachedTest::testIncrementItemInitialValue

Failed asserting that false matches expected 5.

/home/travis/build/zendframework/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:818

3) ZendTest\Cache\Storage\Adapter\MemcachedTest::testIncrementItemsResturnsKeyValuePairsOfWrittenItems

Failed asserting that Array &0 (

    'key1' => 20

) is identical to Array &0 (

    'key1' => 20

    'key2' => 10

).

/home/travis/build/zendframework/zend-cache/test/Storage/Adapter/CommonAdapterTest.php:842

4) ZendTest\Cache\Storage\Adapter\MemcachedTest::testDecrementItemInitialValue

Failed asserting that false matches expected -5.
```

PR #101, opened 2016-05-26
add composer.lock & update travis config

stefanotorresi:
as per zendframework/zendframework#7660

also entails these changes:

- bump minimum php version to 5.6
- add composer scripts entries
- fix travis.yml env variables to not be affected by travis-ci/travis-ci#1444
- fix build on php 7


thomasvargiu:
@stefanotorresi You should disable `APC` and `memcache` with PHP 7.
Can you update your travis settings?

marc-mabe:
@thomasvargui In my opinion this travis error has nothing to do with this PR.

It's related to https://pear.php.net/bugs/bug.php?id=21007 and found the way into the repo by #82, #74 and #52.

The problem now is that a test for a race condition is using different PHP processes and installing apc_bc on PHP 7 requires an extension load order where it fails installed by pecl. On travis there is a hack that adds the extension two times. That worked but PHP triggers a warning on startup but the warning gets be detected now because PHPUnit needs to run this race condition test in an own process.

If we fix the extension install / loading completely this issue should be gone

stefanotorresi:
@thomasvargiu memcache is already disabled by the way ;)

as for the rest, this PR doesn't change anything about extension loading, and @marc-mabe is on it ;)

we can wait for that issue to be resolved and then I'll rebase this

stefanotorresi:
@marc-mabe ping, I see from IRC that you're working on the same changes I'm doing atm :D let me handle this!

marc-mabe:
@stefanotorresi ok I'll wait - You can check my changes here https://github.com/marc-mabe/zend-cache/tree/travis_apc_php7 the only issue I have is that in PHP-7 apc.so doesn't get loaded this way I think because I add it to `.ci/apcu.so` after this file was added

stefanotorresi:
@marc-mabe should be good now!

marc-mabe:
@stefanotorresi  Looks much better now :+1:

But I still have two questions:
* why did you increase the minimum required PHP version? Did I missed something?
* Why do you load apc_bc manuelly instead of using pecl + removing the 'extension=' part from php.ini as in loading apc.so?

weierophinney:
> why did you increase the minimum required PHP version? Did I missed something?

We haven't announced it, but I've talked about it on IRC. Next minor/major versions of all components will be bumped to 5.6, as 5.5 goes EOL next month. @stefanotorresi is making that change along with the composer.lock additions.

marc-mabe:
@weierophinney Thanks for the info!
I added the `BC Break` label because of that but I'll also backport the changes to `master` without the PHP version upgrade as this actually contains a fix for running the tests in travis.

@stefanotorresi I did some additional changes:
* Install lowest, locked and latest versions of `apcu(_bc)`. If we already need to install the extensions we can also install different versions for different runs ;)
  * replace `INSTALL_APCU_BC` with `APCU_BC_PECL_VERSION`
* use the variables `APCU[_BC]_PECL_VERSION` as marker for install and loading ini file conditions
* removed `composer self-update` as this will be done by travis already
* run composer test scripts as `composer run-script test[-coverage] --timeout=XXXX` because I run into a timout issue introduced by composer see https://travis-ci.org/marc-mabe/zend-cache/jobs/133889679#L896

You can take a look into the changes here https://github.com/marc-mabe/zend-cache/blob/hotfix/101/.travis.yml
If you don't have any objections I'll merge it tomorrow

stefanotorresi:
LGTM :)

marc-mabe:
closed by  1a76834 (develop) and d93aaa2 (master)

PR #112, opened 2016-08-16
fixed #107: fixed redis server version test in Redis::internalGetMetadata()

### LINKED ISSUE #107: Redis Adapter checks majorversion with version_compare but RedisResourceManager returns integer
The `RedisResourceManager` always returns integer as version information. Therefore, the check in `Redis` Adapter wont work as expected.

````
    /**
     * Get redis server version
     *
     * @param string $id
     * @return int
     * @throws Exception\RuntimeException
     */
    public function getMajorVersion($id)
    {
        // check resource id and initialize the resource
        $this->getResource($id);

        return (int) $this->resources[$id]['version'];
    }
```

```
            $redisVersion = $this->resourceManager->getMajorVersion($this->resourceId);

            // redis >= 2.8
            // The command 'pttl' returns -2 if the item does not exist
            // and -1 if the item has no associated expire
            if (version_compare($redisVersion, '2.8',  '>=')) {
                $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
                if ($pttl <= -2) {
                    return false;
                }
                $metadata['ttl'] = ($pttl == -1) ? null : $pttl / 1000;

            // redis >= 2.6
            // The command 'pttl' returns -1 if the item does not exist or the item as no associated expire
            } elseif (version_compare($redisVersion, '2.6', '>=')) {
                $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
                if ($pttl <= -1) {
                    if (!$this->internalHasItem($normalizedKey)) {
                        return false;
                    }
                    $metadata['ttl'] = null;
                } else {
                    $metadata['ttl'] = $pttl / 1000;
                }

            // redis >= 2
            // The command 'pttl' is not supported but 'ttl'
            // The command 'ttl' returns 0 if the item does not exist same as if the item is going to be expired
            // NOTE: In case of ttl=0 we return false because the item is going to be expired in a very near future
            //       and then doesn't exist any more
            } elseif (version_compare($redisVersion, '2', '>=')) {
                $ttl = $redis->ttl($this->namespacePrefix . $normalizedKey);
                if ($ttl <= -1) {
                    if (!$this->internalHasItem($normalizedKey)) {
                        return false;
                    }
                    $metadata['ttl'] = null;
                } else {
                    $metadata['ttl'] = $ttl;
                }

            // redis < 2
            // The commands 'pttl' and 'ttl' are not supported
            // but item existence have to be checked
            } elseif (!$this->internalHasItem($normalizedKey)) {
                return false;
            }
```

marc-mabe:
see #107

### DIFF
--- a/src/Storage/Adapter/Redis.php
+++ b/src/Storage/Adapter/Redis.php
@@ -553,7 +553,7 @@ class Redis extends AbstractAdapter implements
         $metadata = [];
 
         try {
-            $redisVersion = $this->resourceManager->getMajorVersion($this->resourceId);
+            $redisVersion = $this->resourceManager->getVersion($this->resourceId);
 
             // redis >= 2.8
             // The command 'pttl' returns -2 if the item does not exist
@@ -565,8 +565,8 @@ class Redis extends AbstractAdapter implements
                 }
                 $metadata['ttl'] = ($pttl == -1) ? null : $pttl / 1000;
 
-            // redis >= 2.6
-            // The command 'pttl' returns -1 if the item does not exist or the item as no associated expire
+            // redis >= 2.6, < 2.8
+            // The command 'pttl' returns -1 if the item does not exist or the item has no associated expire
             } elseif (version_compare($redisVersion, '2.6', '>=')) {
                 $pttl = $redis->pttl($this->namespacePrefix . $normalizedKey);
                 if ($pttl <= -1) {
@@ -578,7 +578,7 @@ class Redis extends AbstractAdapter implements
                     $metadata['ttl'] = $pttl / 1000;
                 }
 
-            // redis >= 2
+            // redis >= 2, < 2.6
             // The command 'pttl' is not supported but 'ttl'
             // The command 'ttl' returns 0 if the item does not exist same as if the item is going to be expired
             // NOTE: In case of ttl=0 we return false because the item is going to be expired in a very near future
--- a/src/Storage/Adapter/RedisResourceManager.php
+++ b/src/Storage/Adapter/RedisResourceManager.php
@@ -41,16 +41,31 @@ class RedisResourceManager
     /**
      * Get redis server version
      *
-     * @param string $id
+     * @param string $resourceId
+     * @return string
+     * @throws Exception\RuntimeException
+     */
+    public function getVersion($resourceId)
+    {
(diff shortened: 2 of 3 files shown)


src/Storage/Adapter/RedisResourceManager.php
```diff
@@ -41,16 +41,31 @@ public function hasResource($id)
...
+        // check resource id and initialize the resource
+        $this->getResource($resourceId);
+
+        return $this->resources[$resourceId]['version'];
+    }
+
+    /**
+     * Get redis major server version
+     *
      * @param string $id
```
boesing: You should rename the parameter in PHPDoc aswell.

src/Storage/Adapter/Redis.php
```diff
@@ -565,8 +565,8 @@ protected function internalGetMetadata(& $normalizedKey)
                 }
                 $metadata['ttl'] = ($pttl == -1) ? null : $pttl / 1000;
 
-            // redis >= 2.6
-            // The command 'pttl' returns -1 if the item does not exist or the item as no associated expire
+            // redis >= 2.6, < 2.8
```
boesing: if less than 2.8, why wont you check this in `elseif` statement?

src/Storage/Adapter/Redis.php
```diff
@@ -578,7 +578,7 @@ protected function internalGetMetadata(& $normalizedKey)
                     $metadata['ttl'] = $pttl / 1000;
                 }
 
-            // redis >= 2
+            // redis >= 2, < 2.6
```
boesing: if less than 2.6, why wont you check this in `elseif` statement?

src/Storage/Adapter/Redis.php
marc-mabe: I don't understand what you mean. It's testing for `if >=2.8 ... elseif >=2.6 ... elseif >=2.0 else`. What's wrong with that ?

boesing: Ah, just didnt get these other `elseif` Statements. My fault.

src/Storage/Adapter/RedisResourceManager.php
marc-mabe: fixed

PR #113, opened 2016-09-13
Move Coverage reporting to latest env

Koopzington:
Requiring php-coveralls results in a full update of all dependencies instead of composer installing the versions mentioned in the composer.lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,13 +67,13 @@ matrix:
     - php: 5.6
       env:
         - DEPS=locked
-        - TEST_COVERAGE=true
         - APCU_PECL_VERSION="apcu-4.0.10"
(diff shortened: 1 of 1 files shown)


marc-mabe:
@Koopzington This will slow down the tests a lot as generating the code coverage and looking up the latest versions both takes additional time and should be spitted over different build jobs

Koopzington:
uhm... what would slow the tests down? It's not like i added some new feature, i merely moved the execution to the environment where it's supposed to be.
Build execution times haven't increased either compared to the one before this one:
https://travis-ci.org/zendframework/zend-cache/builds/159688399
https://travis-ci.org/zendframework/zend-cache/builds/152799417

marc-mabe:
@Koopzington Oh you are right - installing the latest dependencies only takes some seconds more than installing from lock. the biggest overhead comes from code coverage (~2mins more) but we have to do it at least once

PR #116, opened 2016-10-07
Feature/docblock method chaining consistency

tdutrion:
As discussed in [fluent interface return type vs return self #7193](https://github.com/zendframework/zendframework/issues/7193), this changes remove occurrences of class names and self keyword from the DocBlocks

### DIFF
--- a/src/Pattern/AbstractPattern.php
+++ b/src/Pattern/AbstractPattern.php
@@ -22,7 +22,7 @@ abstract class AbstractPattern implements PatternInterface
      * Set pattern options
      *
      * @param  PatternOptions $options
-     * @return AbstractPattern
+     * @return AbstractPattern Provides a fluent interface
      * @throws Exception\InvalidArgumentException
      */
     public function setOptions(PatternOptions $options)
--- a/src/Pattern/CallbackCache.php
+++ b/src/Pattern/CallbackCache.php
@@ -18,7 +18,7 @@ class CallbackCache extends AbstractPattern
      * Set options
      *
      * @param  PatternOptions $options
-     * @return CallbackCache
+     * @return CallbackCache Provides a fluent interface
      * @throws Exception\InvalidArgumentException if missing storage option
      */
     public function setOptions(PatternOptions $options)
--- a/src/Pattern/ClassCache.php
+++ b/src/Pattern/ClassCache.php
@@ -18,7 +18,7 @@ class ClassCache extends CallbackCache
      * Set options
      *
      * @param  PatternOptions $options
-     * @return ClassCache
+     * @return ClassCache Provides a fluent interface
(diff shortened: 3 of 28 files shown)


marc-mabe:
interestingly the test for PHP-7.0 results in segmentation fault https://travis-ci.org/zendframework/zend-cache/jobs/165922572#L9937 :(

tdutrion:
Could it be related to an [update of the apcu version](https://github.com/travis-ci/travis-ci/issues/6687)? I don't use travis much, I'll keep digging, but if you think that's what it is please let me know.

marc-mabe:
Yea this looks very related. I'll try this setup at home to see if it's it.

In case it's because of APCu and unrelated to your changes I'll merge and see if I can provide a simple reproduce script to help the APCu developer.

Anyway thanks for the catch!

marc-mabe:
@tdutrion ok the segfault definitely happens with APCu-5.1.6 -> see https://bugs.php.net/bug.php?id=73286 and krakjoe/apcu#203

sergeyklay:
Yes, just use 5.1.5:
```sh
git clone -q https://github.com/krakjoe/apcu -b v5.1.5 /tmp/apcu
cd /tmp/apcu

phpize &> /dev/null
./configure &> /dev/null

make --silent -j4 &> /dev/null
make --silent install
```

```sh
git clone -q https://github.com/krakjoe/apcu-bc /tmp/apcu-bc
cd /tmp/apcu-bc

phpize &> /dev/null
./configure &> /dev/null

make --silent -j4 &> /dev/null
make --silent install
```

**apcu_bc.ini**
```ini
[APCu]
extension=apcu.so

[APC]
extension=apc.so
apc.enabled=1
apc.enable_cli=1
```

See https://github.com/krakjoe/apcu/issues/203

marc-mabe:
FYI: the APCu issue has been fixed in the latest version

src/Pattern/AbstractPattern.php
```diff
@@ -22,7 +22,7 @@
      * Set pattern options
      *
      * @param  PatternOptions $options
-     * @return AbstractPattern
+     * @return $this Provides a fluent interface
```
DASPRiD: `$this` is not a valid return type, in terms of PHP-doc. Does any IDE actually understand that?

tdutrion: Please go up in my very first message (before all the problems with apcu), and give your opinion (the thread is pretty much dead at the moment...).

I can definitely edit this PR as long as we settle on a comment that can be used throughout the codebase.

tdutrion:
The PR has been edited to reflect the latest outcomes of the discussion :)

tdutrion:
Updated + up to date with the develop branch.

PR #120, opened 2016-10-27
Adding option to change cache and tag file suffixes

Lansoweb:
With this PR we can change the suffix for cache files. One use case is to use the filesystem cache to generate zend-view files (phtml suffixes).


marc-mabe:
Thanks for contributing to zend-cache :)

I understand your use case and it's fine for me even if I think that's a very low used feature but I have one small problem with it:

What happens if the specified suffix contains a special character used by glob?
So you have to escape these characters! PHP doesn't have a function for that but I found one in [python](https://docs.python.org/3/library/glob.html#glob.escape)

Cheers

weierophinney:
@Lansoweb I've rebased and incorporated the feedback from @marc-mabe .

Lansoweb:
@weierophinney Great. Thanks!

weierophinney:
Filesystem adapter unit tests run properly, so going to go ahead and merge to develop.

weierophinney:
Thanks, @Lansoweb!

PR #128, opened 2017-02-07
Fixed incorrect variable usage in MongoDbResourceManager

careys7:
Updated incorrect `$this->resource` variable usages to `$this->resources` in MongoDbResourceManager.

### DIFF
--- a/src/Storage/Adapter/MongoDbResourceManager.php
+++ b/src/Storage/Adapter/MongoDbResourceManager.php
@@ -117,9 +117,9 @@ class MongoDbResourceManager
     {
         $this->resources[$id]['server'] = (string)$server;
 
-        unset($this->resource[$id]['client_instance']);
-        unset($this->resource[$id]['db_instance']);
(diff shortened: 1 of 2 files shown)


Ocramius:
Tests missing. Specifically: the fact that this didn't affect the build status means that we either:

 * simply don't test it
 * had unknown memory leaks
 * it's dead code, and not relevant at all

careys7:
Updated with tests. 

Tests ensure that the instance of `MongoCollection` held for a given resource id does not persist following method calls to:
- `MongoDbResourceManager::setServer()`
- `MongoDbResourceManager::setCollection()`
- `MongoDbResourceManager::setConnectionOptions()`
- `MongoDbResourceManager::setDriverOptions()`
- `MongoDbResourceManager::setDatabase()`

marc-mabe:
@careys7 Thanks for the fix! I have modified you test(s) to get batter coverage.

careys7:
Your tests look better :+1:

Issue #93, opened 2016-03-24
Component split


marc-mabe:
I would like to get your opinions about splitting zend-cache into different components.
The reason for that is simply a testing and dependency hell.

* It's not possible to define a required extension in composer.json so all adapters have to manually check if the current environment is matching the requirements
* Testing against all adapters is a nightmare and takes very long.
   * The travis.yml file al already very complex but it still doesn't test all common cases.
   * Some extensions are missing or not 100% compatible for PHP-7 which makes things here more complicated.
   * Doesn't test against different (most common) extension versions
   * Nearly impossible to run tests for ZendServer adapters
* another reason is that normally you only need 1 or 2 adapters but that's not very important as zend-cache doesn't install so many files per adapter.

In my opinion it makes sense to split parts into it's own repository as long as the part requires a non standard extension or another currently optional dependency.

Thoughts ?

This is a structure how this could look like:

zendframework/zend-cache
-----------------------------------
  * Exceptions
  * Patterns
  * Service
  * Factories
  * Storage Interfaces
  * Storage Plugins
    * all except the Serializer plugin
  * Storage Adapters
    * BlackHole
    * Filesystem
    * Memory

zendframework/zend-cache-serializer
------------------------------------------------
  * adds the serializer storage plugin which requires zend-serializer

zendframework/zend-cache-apc
-----------------------------------------
  * adds the Apc storage adapter requires the apc extension (or compatible extension)

zendframework/zend-cache-apcu
-------------------------------------------
  * adds the Apcu storage adapter requires the apcu extension

zendframework/zend-cache-dba
-----------------------------------
  * adds the Dba storage adapter requires the dba extension

zendframework/zend-cache-memcache
-----------------------------------
  * adds the Memcache storage adapter requires the memcache extension

zendframework/zend-cache-memcached
-----------------------------------
  * adds the Memcached storage adapter requires the memcached extension

zendframework/zend-cache-mongo
-----------------------------------
  * adds the Mongo storage adapter requires the mongo extension
  * It's currently the MongoDb adapter but there is also another extension called mongodb

zendframework/zend-cache-redis
-----------------------------------
  * adds the Redis storage adapter requires the redis extension

zendframework/zend-cache-session
-----------------------------------
  * adds the Session storage adapter requires the zend-session

zendframework/zend-cache-wincache
-----------------------------------
  * adds the WinCache storage adapter requires the wincache extension

zendframework/zend-cache-xcache
-----------------------------------
[...]

PR #129, opened 2017-03-02
phpunit update

kokspflanze:
added phpunit 6 support
travis update

### DIFF
--- a/test/_autoload.php
+++ b/test/_autoload.php
@@ -1,0 +1,12 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+if (!class_exists('PHPUnit_Framework_TestCase')) {
+    class_alias(\PHPUnit\Framework\TestCase::class, 'PHPUnit_Framework_TestCase');
+}
--- a/test/_autoload.php
+++ b/test/_autoload.php
@@ -7,6 +7,8 @@
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
-if (!class_exists('PHPUnit_Framework_TestCase')) {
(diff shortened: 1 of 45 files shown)


kokspflanze:
the error is from https://github.com/php-cache/integration-tests/blob/master/src/CachePoolTest.php#L17

kokspflanze:
that error is fixed with https://github.com/zendframework/zend-eventmanager/pull/47/files

test/_autoload.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+use PHPUnit\Framework\TestCase;
+
+if (! class_exists('PHPUnit_Framework_TestCase')) {
+    class_alias(TestCase::class, 'PHPUnit_Framework_TestCase');
```
webimpress: We don't need this file and this alias here. In PHPUnit 5.7 and 6.0 we already have `TestCase` class.

kokspflanze: the problem is https://github.com/php-cache/integration-tests/blob/master/src/CachePoolTest.php#L17 that is used for the unit-tests, that test is in a external repo and use the old class.

marc-mabe:
Thanks for working on this!
But I'm in the middle of refactoring/splitting the component and will get tons of conflicts afterwards :(
Not sure how to handle the situation best as my time is very limited.

marc-mabe:
See #93

kokspflanze:
@marc-mabe no problem i just want to help with the test problem, if you already have a solution we can close it.

marc-mabe:
I guess we can merge it to master but not to develop as the refactoring will be a new mayor

kokspflanze:
i also need a new mayor, because the phpunit part will just work on php5.6+ and i used base develop, to get the latest changes.

marc-mabe:
Ah I see - didn't noted it on phone.
-> I'll leave your PR open and make your changes manually on refactoring

weierophinney:
Thanks, @kokspflanze!

weierophinney:
Rebased and merged to develop with 1bf8a61.

PR #132, opened 2017-04-05
Use MongoDB instead of legacy mongo

naprstek:
Zend\Cache and Zend\Session uses different support for Mongo as a storage. This change use the same way for Zend\Cache and therefore both modules can be used together.

I could not tun test, it fails on missing functions in another classes (dependency on ZF2?).

Regards
Vladimir

### DIFF
--- a/src/Storage/Adapter/MongoDb.php
+++ b/src/Storage/Adapter/MongoDb.php
@@ -261,7 +261,7 @@ class MongoDb extends AbstractAdapter implements FlushableInterface
      */
     private function fetchFromCollection(& $normalizedKey)
     {
-		try {
+	try {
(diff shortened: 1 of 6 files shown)


weierophinney:
This is a great change, but misses an important aspect of backwards compatibilty: some users are still using ext/mongo.

My feeling is that we should provide separate adapters, keeping the one targeting ext/mongo, and adding a new one targeting ext/mongodb. The latter could be a copy-paste of the former, with the changes you propose in here. Doing so would ensure that current ext/mongo users can continue using zend-cache, while those who are using ext/mongodb can have an adapter that works for them.

I can help out with this, if you like.

weierophinney:
Closing in favor of #154.

PR #134, opened 2017-04-30
Give php the absolute `use` path to Traversable

TysonAndre:
Without that use statement, the check would always be false. See https://3v4l.org/#preview
This was detected by etsy/phan, and the bug may or may not occur in real code.

This change affects setFromArray() on line 315


weierophinney:
Thanks, @TysonAndre

PR #135, opened 2017-07-22
remove .swp file on FileSystem::internalSetItem(s)() when it exists after put content

samsonasik:
I use ubuntu 14.04.4 LTS and when file created, it created ".swp" file like the following: 

```
../
./
zfcache-foo.dat
.swp
```

while I tried in centos, the ".swp" file is not created during "setItem()"

### DIFF
--- a/src/Storage/Adapter/Filesystem.php
+++ b/src/Storage/Adapter/Filesystem.php
@@ -971,7 +971,7 @@ class Filesystem extends AbstractAdapter implements
             $this->putFileContent($filespec . '.dat', $value);
         }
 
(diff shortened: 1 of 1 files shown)


samsonasik:
more specific use case: when we are using `optimizeByFactor` plugin ( call fileSystem->optimize() ), the folder never cleared when the ".swp" file exists and leaving many folders with empty ".swp" file because optimize only remove empty directory.

samsonasik:
after investigating, it is mostly because of when I open  directory with `vim'`, it auto-create ".swp" file on my ubuntu server which when I use normal "cd", it works as expected. closing, I just need to make `optimizing_factor` option of optimizeByFactor low.

PR #146, opened 2018-02-11
Change `@return Exception` to `@throws Exception`

TysonAndre:
In the Dba storage adapter.

Noticed via static analysis


weierophinney:
Thanks, @TysonAndre

PR #141, opened 2017-10-25
PHP 7.2 + QA Tools - follow up #129, #132, #134

webimpress:
Closes #129 
Closes #132 
Closes #134 
Resolves #139 

/cc @froschdesign

PR #148, opened 2018-04-17
Modernize the package and finalize QA tooling

weierophinney:
This patch modernizes the package to follow guidelines as set in our [maintainers repository](https://github.com/zendframework/maintainers).

In particular:

- Moves documentation from `/doc/` to `/docs/`.
- Adds contribution and support docs.
- Normalizes the `composer.json` structure.
- Normalizes various package files (`.gitignore`, `.gitattributes`, etc.)
- Ensures the Travis configuration uses up-to-date practices with regards to lowest/locked/latest strategy.

Additionally, this patch incorporates commits from #141 that accomplish the following:

- Provides a polyfill for the `PatternPluginManager` to ensure it works with both v2 and v3 versions of zend-servicemanager, fixing an issue with PHP 7.2 compatibility.
- Adds PHPUnit 7 as a potential target for running unit tests.
- Provides a number of features around enabling extensions that act as cache backends.
- Drops HHVM support.


weierophinney:
Pinging @webimpress ...

weierophinney:
Initial test runs that have actually completed (a number had segfaults and needed to be restarted) have been mostly successful, with one specific failure: the `Cache\IntegrationTests\CachePoolTest::testHasItemReturnsFalseWhenDeferredItemIsExpired` test case often fails. I suspect this is a timing issue primarily.

weierophinney:
I honestly think at this point the only way we'll get tests to run completely is to separate the adapters into their own repos, as @marc-mabe has planned for some time.

webimpress:
@weierophinney 
> Pinging @webimpress ...

Sorry, I was not able to have a look before... LGTM 👍 😄

thomasvargiu:
@weierophinney About `Cache\IntegrationTests\CachePoolTest::testHasItemReturnsFalseWhenDeferredItemIsExpired` failed test, it's normal.

The `Zend\Cache\Psr\CacheItemPoolAdapter` adapter doesn't support driver deferred saves, so cache items are saved
on destruct or on explicit `commit()` call.

I'm working on it and I'll open a PR to skip that test in a while.

PR #149, opened 2018-04-17
Skip tests for unsupported PSR-6 driver deferred save

thomasvargiu:
The `Zend\Cache\Psr\CacheItemPoolAdapter` adapter doesn't support driver deferred saves, so cache items are saved on destruct or on explicit `commit()` call.

Tests should be skipped.

Please check what I wrote, my english is not good enough.


weierophinney:
Thanks, @thomasvargiu!

thomasvargiu:
@weierophinney I'm preparing another PR fixing acpu and apc minimum versions for tests, then tests should pass

PR #150, opened 2018-04-17
Fixed memcached 3 cas token

thomasvargiu:
[memcached](http://php.net/manual/en/memcached.get.php) >= 3 changed implementation to retrieve the cas token.


weierophinney:
Thanks, @thomasvargiu!

Issue #137, opened 2017-09-25
Redis Adapter: not found server host key


lpj145:
Somebody help me with this ??
https://docs.zendframework.com/zend-cache/storage/adapter/#the-redis-adapter
I try it on config:
```
'cache' => [
        'enabled' => true,
        //About redis configuration
        'redis' => [
            'database' => 0,
            'namespace_separator' => ':',
            'password' => '',
            'persistent_id' => '',
            'resource_manager' => null,
            //This provides redis connector
            //By uri, associative array or list
            'server' => [
                'localhost',
                6379
            ]
        ],
    ]
```
```return new Redis($config['redis']);```

when debug it i found this bug ????

![image](https://user-images.githubusercontent.com/11980109/30832683-5b6aedda-a222-11e7-979c-af1cc5918ec6.png)

**Comment by lpj145** [comment]
Adapter redis not work right ?

**Comment by bcremer** [comment]
This issue seems to be related to https://github.com/zendframework/zend-cache/issues/109#issuecomment-337229452.

The server-key _must_ be first array key in the configuration.

```
'cache' => [
        'enabled' => true,
        //About redis configuration
        'redis' => [
            'server' => [
                'localhost',
                6379
            ]
            'database' => 0,
            'namespace_separator' => ':',
            'password' => '',
            'persistent_id' => '',
            'resource_manager' => null,
            //This provides redis connector
            //By uri, associative array or list

        ],
    ]
```

PR #151, opened 2018-04-18
Redis fixes

### LINKED ISSUE #109: RedisResourceManager breaks if being configured via StorageCacheAbstractServiceFactory/array and service isn't first member of array
I take the example from the docs to demonstrate the problem:

`    'caches' => array(
        'Cache\Transient' => array(
            'adapter' => 'redis',
            'ttl'     => 3600,
            'options' => array(
                'server' => array(
                    'host'=>'localhost',
                    'port' => 6379,
                    'timeout' => 2.5,
                ),
                'database' => 5,
                'password' => '',
            ),
            'plugins' => array(
                'exception_handler' => array(
                    'throw_exceptions' => false,
                ),
                'serializer',
            ),
        ),
    ),`

The relevant part is the options array.

`            'options' => array(
                'server' => array(
                    'host'=>'localhost',
                    'port' => 6379,
                    'timeout' => 2.5,
                ),
                'database' => 0,
                'password' => '',
            ),`
would work, but for example 
`            'options' => array(
                'database' => 0,
                'server' => array(
                    'host'=>'localhost',
                    'port' => 6379,
                    'timeout' => 2.5,
                ),
                'password' => '',
            ),`
wouldn't.

All the setters in RedisResourceManager look if the id is already initialised, if not, every setter calls setResource with only his own parameter.
setResource merges this with the default array, which has empty array for the server key. At the end of the function setResource always calls normalizeServer, which checks if key host is set and throws an exception, if it's not.
As it is now, the code fails for every initialisation, where server key isn't the first inside the options array, because an exception is thrown.

I'm not making own fix via pull request, because this should be a design decision.

thomasvargiu:
Fixed #109, #137
Fixed resource check on RedisResourceManager.
Fixed normalization of libOptions, normalizing them on resource initialization.

### DIFF
--- a/src/Storage/Adapter/RedisOptions.php
+++ b/src/Storage/Adapter/RedisOptions.php
@@ -20,7 +20,7 @@ class RedisOptions extends AdapterOptions
      *
      * @var string[]
      */
-    protected $__prioritizedProperties__ = ['resource_manager', 'resource_id'];
+    protected $__prioritizedProperties__ = ['resource_manager', 'resource_id', 'server'];
     // @codingStandardsIgnoreEnd
 
     /**
--- a/src/Storage/Adapter/RedisResourceManager.php
+++ b/src/Storage/Adapter/RedisResourceManager.php
@@ -147,6 +147,8 @@ class RedisResourceManager
         $resource['resource'] = $redis;
         $this->connect($resource);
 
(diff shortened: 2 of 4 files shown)


weierophinney:
Thanks, @thomasvargiu!

PR #152, opened 2018-04-18
Provide a PSR-16 decorator for zend-cache storage adapters

weierophinney:
This patch serves as an alternative to #126, based on [a comment by @marc-mabe to that patch](https://github.com/zendframework/zend-cache/pull/126#issuecomment-271992110).

The patch adds a new class, `Zend\Cache\Psr\SimpleCacheDecorator`, which accepts a zend-cache storage instance to its constructor, and then proxies to it via the methods that implement the PSR-16 `CacheInterface`. The approach is simpler than that in #126 as it does not require testing against every adapter; we can test the decorator against a mock of the `StorageInterface`, validating its behavior. The tests provided also ensure that any exceptions thrown by the underlying storage operations are re-cast to PSR-16 implementations.

This approach may miss a few details; each of @marc-mabe, @kynx, and @boesing raised a few issues in #126 that I am uncertain if this approach covers. I'd love feedback before I merge this.

#### TODO

- [x] Validate keys to ensure they do not contain reserved characters
- [x] Validate keys to ensure they do not exceed maximum length
- [x] Detect whether the zend-cache adapter requires serialization; if so, serialize values before storage, and deserialize on retrieval.
- [x] Implement TTL capabilities per PSR-16:
  - [x] If a negative or zero TTL is provided, have `set()` and `setMultiple()` remove the item(s) from the cache (instead of storing).
  - [x] If the adapter does not support TTL semantics, have `set()` and `setMultiple()` return boolean false for non-null, `>0` TTL values.
- [ ] Implement integration tests for as many adapters as make sense.

### DIFF
--- a/src/Psr/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCacheDecorator.php
@@ -7,9 +7,13 @@
 
 namespace Zend\Cache\Psr;
 
+use DateInterval;
+use DateTimeImmutable;
+use DateTimeZone;
 use Exception;
 use Psr\SimpleCache\CacheInterface as SimpleCacheInterface;
 use Throwable;
+use Traversable;
 use Zend\Cache\Exception\InvalidArgumentException as ZendCacheInvalidArgumentException;
 use Zend\Cache\Storage\FlushableInterface;
 use Zend\Cache\Storage\StorageInterface;
@@ -24,7 +28,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
     /**
      * Characters reserved by PSR-16 that are not valid in cache keys.
      */
-    const INVALID_KEY_CHARS = '@{}()/\\';
+    const INVALID_KEY_CHARS = ':@{}()/\\';
 
     /**
      * @var bool
@@ -44,11 +48,17 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     private $success;
 
+    /**
+     * @var DateTimeZone
+     */
+    private $utc;
+
     public function __construct(StorageInterface $storage)
     {
         $this->memoizeSerializationCapabilities($storage);
         $this->memoizeTtlCapabilities($storage);
         $this->storage = $storage;
+        $this->utc = new DateTimeZone('UTC');
     }
 
     /**
@@ -56,6 +66,8 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     public function get($key, $default = null)
     {
+        $this->validateKey($key);
+
         $this->success = null;
         try {
             $result = $this->storage->getItem($key, $this->success);
@@ -80,10 +92,10 @@ class SimpleCacheDecorator implements SimpleCacheInterface
     public function set($key, $value, $ttl = null)
     {
         $this->validateKey($key);
+        $ttl = $this->convertTtlToInteger($ttl);
 
         // PSR-16 states that 0 or negative TTL values should result in cache
         // invalidation for the item.
-        $ttl = null !== $ttl ? (int) $ttl : null;
         if (null !== $ttl && 1 > $ttl) {
             return $this->delete($key);
         }
@@ -116,6 +128,12 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     public function delete($key)
     {
+        $this->validateKey($key);
+
+        if (! $this->storage->hasItem($key)) {
+            return true;
+        }
+
         try {
             return $this->storage->removeItem($key);
         } catch (Throwable $e) {
@@ -141,6 +159,9 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     public function getMultiple($keys, $default = null)
     {
+        $keys = $this->convertIterableToArray($keys, false, __FUNCTION__);
+        array_walk($keys, [$this, 'validateKey']);
+
         try {
             $results = $this->storage->getItems($keys);
         } catch (Throwable $e) {
@@ -150,7 +171,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         }
 
         foreach ($keys as $key) {
-            if (! isset($results[$key]) && null !== $default) {
+            if (! isset($results[$key])) {
                 $results[$key] = $default;
                 continue;
             }
@@ -170,7 +191,8 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     public function setMultiple($values, $ttl = null)
     {
-        $ttl = null !== $ttl ? (int) $ttl : null;
+        $values = $this->convertIterableToArray($values, true, __FUNCTION__);
+        $ttl = $this->convertTtlToInteger($ttl);
 
         foreach (array_keys($values) as $key) {
             $this->validateKey($key);
@@ -205,7 +227,18 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         }
 
         $options->setTtl($previousTtl);
-        return $result;
+
+        if (empty($result)) {
+            return true;
+        }
+
+        foreach ($result as $index => $key) {
+            if (! $this->storage->hasItem($key)) {
+                unset($result[$index]);
+            }
+        }
+
+        return empty($result);
     }
 
     /**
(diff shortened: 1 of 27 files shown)


weierophinney:
Pinging @kynx and @boesing for review.

thomasvargiu:
@weierophinney Can we also add this on composer.json?

```json
    "provide": {
        "psr/cache-implementation": "1.0",
        "psr/simple-cache-implementation": "1.0"
    }
```

weierophinney:
@thomasvargiu Done!

src/Psr/SimpleCacheDecorator.php
```diff
@@ -0,0 +1,187 @@
...
+            throw static::translateException($e);
+        } catch (Exception $e) {
+            throw static::translateException($e);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function set($key, $value, $ttl = null)
```
kynx: Yeah, same here

kynx: Oops darn mobile interface are my comment: the spec forbids some character (like @) in the key. These should be caught early.

weierophinney: Oh, okay - I'll review the spec and add some checks here and in `setMultiple()` then.

weierophinney: I've added a validation method that does the following:

- Checks for any characters reserved by the PSR-16 specification, raising an exception if any are detected.
- Checks the key length; anything over 64 characters results in an exception.

Both `set()` and `setMultiple()` now call this before doing any other logic.

weierophinney:
> The thing that made the PSR-6 adapter such a pain was handling the different limitations of the underlying storage and the different values they can store.
> 
> I’m wondering if it would be better to serialise everything up front to avoid those.

I just turned to that section of the specification, and it looks like it may be identical to PSR-6. As such, I'm going to implement the logic found in the PSR-6 adapter with regards to:

- determining if serialization is necessary
- de/serializing values during set/get(Multiple) operations

The only other consideration is what to do about adapters that do not support per-item TTL, as this is a soft requirement of PSR-16 (see: requirements on `set()` and `setMultiple()` with regards to `$ttl` arguments). If no `$ttl` argument is sent, then _any_ adapter is valid. However, if one _is_ sent, PSR-16 is clear that the adapter should honor it.

The options I see are:

- We raise an exception during instantiation if the adapter cannot support TTL, like we do in the PSR-16 adapter. The downside of this approach is that if the consumer is not using TTL values, then they are excluded from using that adapter for no reason.
- We return boolean `false` from the `set()` and `setItems()` operations if a `$ttl` is provided, but the adapter is not capable of it. A boolean `false` indicates a failure to cache, which would be valid in cases where we cannot set a TTL.

Considering the idea behind PSR-16 was to make caching simple and to prevent raising exceptions for common operations, I think the second approach likely would be more in the spirit of PSR-16 support, so I'm going to proceed in that direction.

src/Psr/SimpleCacheDecorator.php
```diff
@@ -0,0 +1,220 @@
...
+    {
+        $this->validateKey($key);
+        $options = $this->storage->getOptions();
+        $previousTtl = $options->getTtl();
+        $options->setTtl($ttl);
+
+        try {
+            $result = $this->storage->setItem($key, $value);
```
thomasvargiu: > If a negative or zero TTL is provided, the item MUST be deleted from the cache if it exists, as it is expired already.

Should it be checked?

weierophinney: Good catch; I'll address this when working on the TTL items I listed above.

src/Psr/SimpleCacheDecorator.php
```diff
@@ -0,0 +1,220 @@
...
+     * @param string $key
+     * @return void
+     * @throws SimpleCacheInvalidArgumentException if key is invalid
+     */
+    private function validateKey($key)
+    {
+        $regex = sprintf('/[%s]/', preg_quote(self::INVALID_KEY_CHARS, '/'));
+        if (preg_match($regex, $key)) {
+            throw new SimpleCacheInvalidArgumentException(sprintf(
+                'Invalid key "%s" provided; cannot contain any of ()',
```
boesing: I guess you wanted to add a %s inside the braces with an implode of self::INVALID_KEY_CHARS?

weierophinney: Good catch - thanks!

thomasvargiu:
@weierophinney Did you consider to test it using [php-cache/integration-tests](https://github.com/php-cache/integration-tests/blob/master/src/SimpleCacheTest.php) like PSR-6 tests?

weierophinney:
> Did you consider to test it using php-cache/integration-tests like PSR-6 tests?

I'll look into it today. I'm not sure if this will be terribly feasible, as we'd need an actual adapter (not a mock one) to easily use those. It may make more sense to wait to do those until we split the adapters into their own repos.

weierophinney:
@thomasvargiu —

Regarding this statement:

> I'll look into it today. I'm not sure if this will be terribly feasible, as we'd need an actual adapter (not a mock one) to easily use those. It may make more sense to wait to do those until we split the adapters into their own repos.

I've added an integration test for the APC adapter, and discovered a number of things already. One thing interesting: the integration tests are sometimes covering behavior not explicitly detailed in the spec. As examples: they test for generator/iterable arguments to the various `*Multiple()` methods, which are not explicitly indicated as valid values, but are not _excluded_, either. Additionally, I'd forgottent that `DateInterval` values were allowed for TTL values, so I've added functionality for addressing that, too.

As such, I think we definitely need _some_ integration tests in play, as they're helping me refine the adapter. I'll get these in place today.

thomasvargiu:
@weierophinney good, and I agree with you. I think should be better to have an integration test for every adapter like in PSR-6 tests. Then we can decide what to do when some adapter fail.
I don't know what should be better to do when someone sets a TTL on some adapter that doesn't support it.

weierophinney:
Initial tests are now in place for APC and APCu adapters; APCu was running fine locally at this point, which is promising!

I also found a bug in the integration tests and reported it: https://github.com/php-cache/integration-tests/issues/92

I'll hammer through the tests for the remaining adapters this afternoon; they should be fairly straight-forward, as I can use the setup logic from the PSR-6 integration tests.

weierophinney:
Looks like the bulk of integration test issues are when using setMultiple() with adapters that require serialization; essentially every single one that requires serialization had `unserialize()` errors, which makes me think that the logic for `setMultiple()` may need tweaking.

src/Psr/SimpleCacheDecorator.php
```diff
@@ -0,0 +1,187 @@
...
+
+        try {
+            $result = $this->storage->setItems($values);
+        } catch (Throwable $e) {
+            throw static::translateException($e);
+        } catch (Exception $e) {
+            throw static::translateException($e);
+        }
+
+        $options->setTtl($previousTtl);
```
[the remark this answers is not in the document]
weierophinney: Done!

weierophinney:
Okay, all feedback incorporated, and no changes at this time that affect tests, which were green on last run!

Going to merge to develop, and will release 2.8.0 next Monday. Please test, and provide feedback if you run into issues!

marc-mabe:
Hi @weierophinney,

Thanks for starting working at it!

* there is a serializer plugin available that can be plugged in into any storage adapters. Maybe it’s better to throw an error if the provided storage adapter doesn’t support all required data types and let the user plugin the serializer plugin into it to make it work with the psr wrapper. I think it was done that way for psr6, too
* also now there is one namespace for both psr wrappers. I think it could make sense to plit it into psr6 and psr16 namespaces 
* on clearing items there is also a ‘clearByNamespaceInterface’ that will remove all items of the current namespace of the storage. As the simple cache wrapper doesn’t has any knowledge about namespaces it only sees the items of the current namespace. I didn’t check the spec fully but I guess this should be preferred on clear as flush removes everything even of other namespaces
* on delete there is a concurrent regression. You first check for existence and then delete but the item could be added between both calls. As removeItem return true/false based of if the item was deleted it will throw an error only in case the deletion of an existing item failed. So probably just ignoring the return value and return true in all cases

weierophinney:
@marc-mabe Thanks for the feedback!

I've opened #155 to address your second point; however, the approach I took was to use `Zend\Cache\Psr\CacheItemPool` and `Zend\Cache\Psr\SimpleCache` to reflect the interfaces being implemented, versus the specification numbers.

I'll start addressing the other three points in the next few days. I'm going to open tickets for each so that others can more easily find the information.

PR #153, opened 2018-04-18
Fixed minimum dependencies and test mongo adapter on php 7.x

thomasvargiu:
This PR fix minimum required apcu and apc_bc versions. Some versions of those cause segmentation faults.


thomasvargiu:
@weierophinney Done! Everything is green again (finally)!

weierophinney:
Thanks, @thomasvargiu!

PR #160, opened 2018-04-22
WIP: PSR-16 Return true deleting a key that was no set

### LINKED ISSUE #158: PSR-16 support: prevent concurrent regression when calling delete
Per [a comment](https://github.com/zendframework/zend-cache/pull/152#issuecomment-383011006) by @marc-mabe : 

> on delete there is a concurrent regression. You first check for existence and then delete but the item could be added between both calls. As removeItem return true/false based of if the item was deleted it will throw an error only in case the deletion of an existing item failed. So probably just ignoring the return value and return true in all cases

thomasvargiu:
Fix #158

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -57,7 +57,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
 
     public function __construct(StorageInterface $storage)
     {
-        $this->attachExceptionHandlers();
+        $this->attachExceptionHandlers($storage);
         $this->memoizeSerializationCapabilities($storage);
         $this->memoizeTtlCapabilities($storage);
         $this->storage = $storage;
@@ -73,9 +73,9 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      *
      * Listeners are attached with the lower priority so the ExceptionHandler plugin is able to throw every exception.
      */
-    private function attachExceptionHandlers()
+    private function attachExceptionHandlers(StorageInterface $storage)
     {
-        if (! $this->storage instanceof EventsCapableInterface) {
+        if (! $storage instanceof EventsCapableInterface) {
             return;
         }
 
@@ -85,7 +85,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         ];
 
         foreach ($events as $event) {
-            $this->storage->getEventManager()->attach(
+            $storage->getEventManager()->attach(
                 $event,
                 function ($e) {
                     throw new StorageException('A storage exception occurred', 0, $e);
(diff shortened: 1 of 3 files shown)


marc-mabe:
@thomasvargiu I don’t think it’s a good idea to modify the injected storage object. I can see what you are trying to solve and I also don’t have a good idea but in my opinion it should be more safe to document the behavior if the exception plugin is used that it’s not psr confirm.

Or as alternative we could add another capability to export the knowledge if exception will be thrown on failure and change that behavior in exception handler plugin. Then the psr decorator can ask for it and throw exception on constructor

thomasvargiu:
@marc-mabe You're right, it's no a good idea, the adapter can be used somewhere else and we are changing its behavior.

The first idea was to use the event manager, but using a variable by reference. The problem is I need to   attach and detach the listener for every call. What do you think about?

weierophinney:
I think any approach that requires attaching or detaching listeners within the decorator is a no-go, as doing so has side effects. If used in an async system where the adapter is shared between requests, we run into an all-too-real possibility that the change could affect usage in a parallel request.

My inclination is as follows:

- We already have a try/catch block. By default, ExceptionEvent instances mark the `throwException` flag as true, which means users will have needed to disable that flag via a configuration setting and usage of the ExceptionHandler plugin, or have a custom plugin in place for handling exceptions. This means that for default usage, the exception will be thrown anyways, we handle it and — and this is where the patch needs to change — we return boolean `false` (the specification for `delete()` says of the return value that it should be "False if there was an error").
- In all other events, we should return the return value of `removeItem()` and/or whether the return value of `removeItems()` is an empty array.
- We should DOCUMENT that the behavior of these methods is somewhat dependent on whether or not the user is using the ExceptionHandler and configured it to NOT throw exceptions, or has another plugin enabled that will intercept exceptions. That documentation should indicate that if any of those situations is true, the return values of `delete()` and `deleteMultiple()` may be false positives.

thomasvargiu:
Are you suggesting to use the same behaviour of adapters (returns `false` when an item doesn't exist)?
PSR-6 decorator returns always `true` even if a key doesn't exist, and looking around all major cache libraries do the same thing.

We should just understand what the PSR specification means on:

> `@return bool True if the item was successfully removed. False if there was an error.`

Is trying to remove a key that doesn't exist an error? Maybe.

Another thing to think about is that there is a `Psr\SimpleCache\CacheException` where the specification say:

> Interface used for all types of exceptions thrown by the implementing library

We can't change storage beahvior because it will result in a BC (maybe in 3.0?) but we should decide at least which beahviour to choose on these decorators.

Another thing to think about is that some adapters have the same problem described in #158 (`Dba` and `Filesystem` at least).
The only concurrent regression that will result in a problem is not when a key is added between the two calls (`hasItem()` and `removeItem()`), but when a key exists and then it's removed between the two calls, returning `false` instead of `true`. We don't mind if an item is added after `hasItem()` because the beahvior will be the same calling directly `removeItem()` I think.
Honestly, I think I would prefer what is currently implemented in the [`develop` branch](https://github.com/zendframework/zend-cache/blob/11518ae6cf6027cb17feb10657434bdf1954bd76/src/Psr/SimpleCache/SimpleCacheDecorator.php#L134).

weierophinney:
@thomasvargiu I've created a new pull request, #163, that I think will handle this correctly; could you review?

thomasvargiu:
Closing in favor of #163

weierophinney:
Superceded by #163

PR #163, opened 2018-04-23
Handle item removal consistent with PSR-16 from SimpleCacheDecorator

### LINKED ISSUE #158: PSR-16 support: prevent concurrent regression when calling delete

weierophinney:
PSR-16 states that `delete()` and `deleteMultiple()` should return false in the case of an error, not throw an exception. This patch modifies the catch statements of each to `return false` accordingly.

Additionally, it modifies the `return $this->storage->removeItem($key)` to read `return (bool) $this->storage->removeItem($key);`. When using the `ExceptionHandler` plugin, it's possible for a `void` or `null` return value due to never receiving a result from the adapter for the operation; by casting to boolean, we ensure it properly indicates failure per the PSR-16 specification.

This patch also documents pitfalls with `delete()` and `deleteMultiple()` with regards to plugins changing the throw behavior and/or altering the result.

Finally, this patch documents which adapters require the Serializer plugin when used with PSR-16.

Alternative to #160 
Fixes #158

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -135,12 +135,8 @@ class SimpleCacheDecorator implements SimpleCacheInterface
     {
         $this->validateKey($key);
 
-        if (! $this->storage->hasItem($key)) {
-            return true;
-        }
-
         try {
-            return (bool) $this->storage->removeItem($key);
+            return null !== $this->storage->removeItem($key);
         } catch (Throwable $e) {
             return false;
         } catch (Exception $e) {
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -142,9 +142,9 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         try {
             return $this->storage->removeItem($key);
         } catch (Throwable $e) {
-            throw static::translateException($e);
+            return false;
         } catch (Exception $e) {
-            throw static::translateException($e);
+            return false;
         }
     }
 
@@ -257,9 +257,9 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         try {
             $result = $this->storage->removeItems($keys);
         } catch (Throwable $e) {
-            throw static::translateException($e);
+            return false;
         } catch (Exception $e) {
-            throw static::translateException($e);
+            return false;
         }
 
         if (empty($result)) {
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -140,7 +140,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         }
 
         try {
-            return $this->storage->removeItem($key);
+            return (bool) $this->storage->removeItem($key);
         } catch (Throwable $e) {
             return false;
         } catch (Exception $e) {
--- a/test/Psr/SimpleCache/SimpleCacheDecoratorTest.php
+++ b/test/Psr/SimpleCache/SimpleCacheDecoratorTest.php
@@ -358,20 +358,13 @@ class SimpleCacheDecoratorTest extends TestCase
         $this->assertTrue($this->cache->delete('key'));
     }
 
-    public function testDeleteShouldReRaiseExceptionThrownByStorage()
+    public function testDeleteShouldReturnFalseWhenExceptionThrownByStorage()
     {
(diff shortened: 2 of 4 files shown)


src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -135,16 +135,12 @@ public function delete($key)
     {
         $this->validateKey($key);
 
-        if (! $this->storage->hasItem($key)) {
-            return true;
-        }
-
         try {
-            return $this->storage->removeItem($key);
+            return null !== $this->storage->removeItem($key);
```
weierophinney: `null` _should_ only happen if an exception was caught, as `$result` is undefined at that point, meaning a `null` value will be returned. Boolean return values indicate that the key does not exist — `true` indicates successful removal and/or key never existed, while `false` generally only means the key wasn't present in the first place. As such, both are successful operations in terms of PSR-16, as they mean the key is not present in the storage. `null` indicates an error, which, by PSR-16, should be indicated via a `false` value.

src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -135,16 +135,12 @@ public function delete($key)
...
         } catch (Throwable $e) {
-            throw static::translateException($e);
+            return false;
         } catch (Exception $e) {
-            throw static::translateException($e);
+            return false;
```
weierophinney: PSR-16 indicates `false` should be returned from `delete()` in the event of an error. An exception is an error condition.

test/Psr/SimpleCache/SimpleCacheDecoratorTest.php
```diff
@@ -346,32 +346,22 @@ public function testSetShouldReRaiseExceptionThrownByStorage()
...
-        $this->storage->hasItem('key')->willReturn(true);
         $this->storage->removeItem('key')->willReturn(true);
         $this->assertTrue($this->cache->delete('key'));
     }
 
     public function testDeleteShouldReturnTrueWhenItemDoesNotExist()
     {
-        $this->storage->hasItem('key')->willReturn(false);
-        $this->storage->removeItem('key')->shouldNotBeCalled();
+        $this->storage->removeItem('key')->willReturn(false);
```
weierophinney: This demonstrates the "`false` returned because key did not exist" scenario.

src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -140,11 +140,11 @@ public function delete($key)
         }
 
         try {
-            return $this->storage->removeItem($key);
+            return (bool) $this->storage->removeItem($key);
```
thomasvargiu: It doesn't fix #158 as a concurrent call can remove a key between `hasItem()` and `removeItem()`, resulting on returning false for a non existing key (against PSR-16 specification). Honestly I don't know what should be better.
Probably returning always `true` as @marc-mabe initially suggested could be better. Right now should be better as we are documenting it saying to don't disable exceptions. (I should read the documentation for the return value of each adapter)

src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -257,9 +257,9 @@ public function deleteMultiple($keys)
         try {
             $result = $this->storage->removeItems($keys);
```
thomasvargiu: If `delete()` should return `true` on a non set item, probably also `deleteMultiple()` should return `true` (??)

src/Psr/SimpleCache/SimpleCacheDecorator.php
thomasvargiu: It's perfect, it was so simple

src/Psr/SimpleCache/SimpleCacheDecorator.php
thomasvargiu: Can we change it with the following?

```php
return null !== $this->storage->removeItems($keys);
```

weierophinney:
Test failure was due to a timeout. I've tried two different ways to increase it, but none of them are being honored, so at this point, we just have to restart jobs as necessary. Since only one environment timed out and all others passed, this is ready.

thomasvargiu:
@weierophinney can we ignore composer scripts and run phpunit directly?

PR #164, opened 2018-04-23
Return false on an exception deleting items

thomasvargiu:
Same of #163 but for PSR-6

### DIFF
--- a/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
+++ b/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
@@ -189,20 +189,16 @@ class CacheItemPoolDecorator implements CacheItemPoolInterface
     {
         $this->validateKeys($keys);
 
-        $deleted = true;
-
         // remove deferred items first
         $this->deferred = array_diff_key($this->deferred, array_flip($keys));
 
         try {
-            $this->storage->removeItems($keys);
+            return null !== $this->storage->removeItems($keys);
         } catch (Exception\InvalidArgumentException $e) {
             throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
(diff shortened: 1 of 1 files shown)


src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -189,20 +189,16 @@ public function deleteItems(array $keys)
...
         // remove deferred items first
         $this->deferred = array_diff_key($this->deferred, array_flip($keys));
 
         try {
-            $this->storage->removeItems($keys);
+            return null !== $this->storage->removeItems($keys);
         } catch (Exception\InvalidArgumentException $e) {
             throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
         } catch (Exception\ExceptionInterface $e) {
-            $deleted = false;
```
samsonasik: I think it can use empty body inside `catch` and return false at the end of function body:

```php
} catch (Exception\ExceptionInterface $e) {}

return false;
```
so it clear that it return bool.

thomasvargiu: Personally i don't like empty catch blocks. Initially I was adding a comment line saying "ignored", but then I realized it was the only exit condition, that's what I wanted to make it clear returning `false` inside the catch block.

samsonasik: I was reading @weierophinney 's post at https://mwop.net/blog/2017-03-30-anonymous-class-middleware.html which uses empty catch when it can be moved to end of function.

thomasvargiu: Done! Thank you @samsonasik

Issue #165, opened 2018-04-26
ZendCache 2.8 sets $success on getItem if there is a null response


dpauli:
- [X] I was not able to find an [open](https://github.com/zendframework/zend-cache/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-cache/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/).)

If you use the `getItem` function from `Storage/Adapter/AbstractAdapter` you have the `$success` reference to check whether it succeed or not. This attribute is also `true` if the `getItem` responseded `null` (getting the item failed)`.

In previous ZendCache 2.7 it works perfectly.

### Code to reproduce the issue

```php
$item = $this->cache->getItem('someCacheKey', $success);

# $success is true but the $item is null
```

### Expected results

The `$success` attribute is needed to know whether a `$item` is found or not, so it should be false.

weierophinney:
Fixed with #166/#167

PR #166, opened 2018-04-26
#165 fix success bool

dpauli:
This fix solve the problem if there could not find a memcache item the `$success` attribute is set correctly to `false` (see #165).

### DIFF
--- a/src/Storage/Adapter/Memcached.php
+++ b/src/Storage/Adapter/Memcached.php
@@ -206,8 +206,8 @@ class Memcached extends AbstractAdapter implements
         if (func_num_args() > 2) {
             if (defined('Memcached::GET_EXTENDED')) {
                 $output = $memc->get($internalKey, null, \Memcached::GET_EXTENDED);
-                $casToken = $output['cas'];
-                $result = $output['value'] ?? false;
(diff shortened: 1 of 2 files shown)


src/Storage/Adapter/Memcached.php
```diff
@@ -207,7 +207,7 @@ protected function internalGetItem(& $normalizedKey, & $success = null, & $casTo
             if (defined('Memcached::GET_EXTENDED')) {
                 $output = $memc->get($internalKey, null, \Memcached::GET_EXTENDED);
                 $casToken = $output['cas'];
-                $result = $output['value'];
+                $result = $output['value'] ?? false;
```
Ocramius: A test is needed here

webimpress: ... and we should use `?:` instead of `??` as repo still supports PHP 5.6 and value `$output['value']` is 'always' set.

dpauli:
Using PHP 5.6 is totally right, sorry for this.

Writing test for this lib means install locally a PHP + PHPUnit. I dont have time to do this right now, I only fixed a critical bug. If you dont want this fixed bug close this PR. :)

thomasvargiu:
I opened a new PR #167 adding tests

PR #167, opened 2018-04-26
Hotfix #165

### LINKED ISSUE #165: ZendCache 2.8 sets $success on getItem if there is a null response

thomasvargiu:
Fix #165.
In favor of #166

### DIFF
--- a/test/Storage/Adapter/MemcachedTest.php
+++ b/test/Storage/Adapter/MemcachedTest.php
@@ -9,6 +9,7 @@
 
 namespace ZendTest\Cache\Storage\Adapter;
 
+use Prophecy\Argument;
 use Zend\Cache;
 
 /**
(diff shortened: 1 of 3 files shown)


test/Storage/Adapter/MemcachedTest.php
```diff
@@ -86,6 +87,33 @@ public function testOptionsAddServer()
...
+            'resource_manager' => $resourceManager->reveal(),
+        ]);
+
+        $storage->getEventManager()->attach(
+            'getItem.exception',
+            function(Cache\Storage\ExceptionEvent $e) {
+                $e->setThrowException(false);
+                $e->stopPropagation(true);
+            },
+            -1);
```
webimpress: Just CS - closing bracket should be in new line in multiline function call.

PR #173, opened 2018-07-02
change substr with strpos/stripos when possible

samsonasik:
- [x] Is this related to quality assurance?


webimpress:
Thanks, @samsonasik !

PR #178, opened 2018-12-26
Added PHP 7.3 support - updated Travis CI configuration

webimpress:


pimjansen:
This is the last one for the 2.9 release right?

webimpress:
@pimjansen Yes, just working on it. Do you have anything else in your mind which should go immediately?

pimjansen:
@webimpress no was just wondering. Since there was more stuff planned for the minor before. Happy that the blocker is going as an in between so good work 👌🏻

Issue #183, opened 2019-03-29
SimpleCacheDecorator Cache gets deleted after creation


alexz707:
I tried to implement a Cache (FileSystem Adapter) and use it with SimpleCacheDecorator.
The problem is that the cache dir will be created but no caches files are in the directory.

After some debugging I have seen that the cache file will be created and after that will be deleted immediately. 
The ttl which I have set in the config will be overwritten by null short before the cache entry is written. So the adapter writes it and clears it because the ttl is null.
After that the ttl will be set back to the config value.

 This happens in the method set (L:97) -> L 114 - 116

My problem gets fixed with the following code:
   
```
if ($ttl !== \null) {
        $options->setTtl($ttl);
}
```
instead of only

`   $options->setTtl($ttl);`
        
Regards 
Alex

PR will be attached shortly

PR #184, opened 2019-03-29
Fixes cache entry deletion directly after creation

alexz707:
See
https://github.com/zendframework/zend-cache/issues/183

Only set TTL if there is a value which is not null!

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -216,7 +216,10 @@ class SimpleCacheDecorator implements SimpleCacheInterface
 
         $options = $this->storage->getOptions();
         $previousTtl = $options->getTtl();
-        $options->setTtl($ttl);
+        
+        if (\null !== $ttl) {
+            $options->setTtl($ttl);
+        }
 
         try {
             $result = $this->storage->setItems($values);
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -114,7 +114,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         $options = $this->storage->getOptions();
         $previousTtl = $options->getTtl();
 
-        if (\null !== $ttl) {
+        if ($ttl !== null) {
             $options->setTtl($ttl);
         }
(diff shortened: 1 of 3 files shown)


src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -113,7 +113,10 @@ public function set($key, $value, $ttl = null)
 
         $options = $this->storage->getOptions();
         $previousTtl = $options->getTtl();
-        $options->setTtl($ttl);
+
+        if (\null !== $ttl) {
```
weierophinney: There's no need to globally qualify `null`; it's an internal type.

webimpress:
@weierophinney I've added tests to cover changes and prove the issue. Instead of mocking storage and options, I've implemented storage and logged TTL when setting the item in cache. Then in assertion I am checking what value of TTL was set for the item.

webimpress:
Actually using mocking it will be also quite simply:

```php
    public function testUseTtlFromOptionsOnSetMocking()
    {
        $this->options->getTtl()->willReturn(40);
        $this->options->setTtl(40)->will([$this->options, 'reveal']);

        $this->options->setTtl(null)->shouldNotBeCalled();

        $this->storage->getOptions()->will([$this->options, 'reveal']);
        $this->storage->setItem('foo', 'bar')->willReturn(true);

        self::assertTrue($this->cache->set('foo', 'bar'));
    }

    public function testUseTtlFromOptionsOnSetMultipleMocking()
    {
        $this->options->getTtl()->willReturn(40);
        $this->options->setTtl(40)->will([$this->options, 'reveal']);

        $this->options->setTtl(null)->shouldNotBeCalled();

        $this->storage->getOptions()->will([$this->options, 'reveal']);
        $this->storage->setItems(['foo' => 'bar', 'boo' => 'baz'])->willReturn([]);

        self::assertTrue($this->cache->setMultiple(['foo' => 'bar', 'boo' => 'baz']));
    }
```

But not sure if it is clear enough?

Line
```php
$this->options->setTtl(null)->shouldNotBeCalled();
```

in both tests is not really needed. It says explicitly that we do not allow that cal, but as we are mocking options anyway, we are saying what calls are allowed (and all others are not).

webimpress:
Thanks, @alexz707!

PR #189, opened 2019-08-15
Update Travis CI config to fix memcache/memcached tests on PHP 5.6

webimpress:
It looks like memcache/memcached services are no longer installed properly on Travis when using PHP 5.6, so we need to use pecl command to install them.

- [x] Is this related to quality assurance?


froschdesign:
Results for 5.6 are all green. 👍

